PR #2, opened 2017-08-02
Update Shopify template to include verification token

unixcharles:
Quite often our customers will want to move domain across shop. We need a way to verify their claim of ownership.


unixcharles:
thanks!

PR #4, opened 2017-09-14
Added the Pro Sites site template

BogdanHabic:
When using the prosites.godaddy.com One Click Migration feature it would greatly simplify the DNS configuration step for our users if we could do it for them semi-automatically.

This template adds what is needed in order for this new flow to work.


BogdanHabic:
Closing this pull request because this template isn't needed anymore.

PR #11, opened 2018-03-07
unbounce template for domain connect

KieranCormack:
Create template for Unbounce to use Domain Connect. 

Creates CNAME record pointing at unbounce pages and accepts a host variable to set sub domains on root domain.

### DIFF
--- a/unbounce.com.website.json
+++ b/unbounce.com.website.json
@@ -1,0 +1,17 @@
+{
+   "providerId":"app.unbounce.com",
+   "providerName": "Unbounce",
+   "serviceId":"site",
+   "serviceName":"Unbounce Public Web Application",
+   "logoUrl":"https://app.unbounce.com/app_assets/sign_up_funnel/unbounce-logo-2017-6c9556be74767c10c9992c13898b84f1.svg",
+   "description":"Enables a Domain to work with Unbounce",
+   "variableDescription":"HOST is the subdomain for the domain.",
+   "records":[
+      {
+        "type": "CNAME",
+        "host": "%HOST%",
+        "pointsTo": "unbouncepages.com",
+        "ttl": "600"
+       }
+   ]
+}


unbounce.com.website.json
```diff
@@ -0,0 +1,17 @@
...
+   "providerName": "Unbounce",
+   "serviceId":"site",
+   "serviceName":"Unbounce Public Web Application",
+   "logoUrl":"https://app.unbounce.com/app_assets/sign_up_funnel/unbounce-logo-2017-6c9556be74767c10c9992c13898b84f1.svg",
+   "description":"Enables a Domain to work with Unbounce",
+   "variableDescription":"HOST is the subdomain for the domain.",
+   "records":[
+      {
+        "type": "CNAME",
+        "host": "%HOST%",
```
arnoldblinn: Recommend using @ as this value, allowing the template to be applied to multiple unique domain/host tuples without conflicts being detected or templates being removed when applying multiple times.

See section 6.4 in spec for explanation.

pawel-kow: Hi @arnoldblinn , it's interesting remark, as maybe we have different understanding of the spec here, so worth clarify
My understanding was that records in the template are always scoped to the domain&host.

So using this example: %HOST% = "secretkey", domain="foo.com", host="" would render secretkey.foo.com CNAME unbouncepages.com

When template applied again with %HOST% = "secretkey", domain="foo.com", host="bar" will render secretkey.bar.foo.com CNAME unbouncepages.com
And no conflict nor replacement of the first template application.

At least this was our understanding and implementation so far.

KieranCormack: @arnoldblinn thanks for the comment.  I was also under the same impression as @pawel-kow can you clarify?

arnoldblinn: I gave this a great deal of thought.

And I realize now a couple of things. In many ways this format and mine are logically equivalent. However, this syntax is a bit more "DNS Friendly".  Having a @ in a CNAME could be confusing.  Where this is clear.

The real concern is variables in hosts in the template. Variables in the host break a couple of things for the DNS Providers. Mainly conflict detection, and template removal. In fact, the notion of the host= parameter on the query string (and hosts= for async permission) was introduced because of this.

But I realize if we modify the rule of "No variables in the host in the template" to "No variables in the host in the template except the past in host value", this approach is cleaner.

PR #13, opened 2018-03-12
update cname to @ to match proposed use of template

KieranCormack:
Using the @ for the host.  We will be sending the sub domain in the domain parameter


arnoldblinn:
So we met on this, and think the %host% syntax is the best approach for this.  The template in the master reflects this.

PR #14, opened 2018-03-12
WP Engine a record and cname templates

markkelnar:
Initial version of the WP Engine templates.

cc @rohjay @brianandrade-wpe


markkelnar:
Thanks @arnoldblinn.  I've made these unique now.

PR #20, opened 2018-05-22
Prevent Mixed Content issue for logoUrl

xgin:
Now we are getting the "Mixed Content" warning when Service Provider logo is displayed.
Since the domainconnect.org has a valid certificate, I propose to switch all URLs to HTTPS.

### DIFF
--- a/about.me.website.json
+++ b/about.me.website.json
@@ -3,7 +3,7 @@
    "providerName":"About Me",
    "serviceId":"website",
    "serviceName":"Hosting",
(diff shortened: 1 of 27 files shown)


arnoldblinn:
I'm OK with this, but bear in mind that my intent of the logos was that this is largely informational and a repository.

If I were implementing a DNS Provider, I certainly wouldn't be pulling logos and rendering them on my site from domainconnect.org.

I would copy the logos to my server.

I suppose we could come up with a better way to do this based on github....maybe each service provider ought to have a logo sourced there?

PR #34, opened 2018-06-04
Rename unbounce.com.website.json to app.unbounce.com.website.json

pawel-kow:
File name changed to match convention.


arnoldblinn:
I would like to do this change; but unbounce is live with us as app.unbounce.com.

So we need to coordinate this a bit....

pawel-kow:
Sure, that's why I renamed the file not the other way around with change of providerId. 
Shall not break a thing.
Same issue is there with wordpress.com.g-suite.json, shall be named wordpress.com.wordpress-services.json

arnoldblinn:
Didn't catch that.  I'm good.  As for the wordpress one, I have a pending mail out to Del to see if we can scrap this template in favor of a shared one for everyone.

PR #36, opened 2018-06-04
Bushlack shopmatic

bushlack:
Removing obsolete shopmatic template and adding a new one for future integration with shopmatic's multiple entities.


myshopmatic.com.website.json
```diff
@@ -1,21 +1,21 @@
 {  
-   "providerId":"goshopmatic.hk",
+   "providerId":"goshopmatic.com",
```
arnoldblinn: The file should be called goshopmatic.com.website.json with this providerId

myshopmatic.com.website.json
```diff
@@ -1,21 +1,21 @@
...
       {  
          "type":"CNAME",
          "host":"www",
-         "pointsTo":"%v2%",
+         "pointsTo":"%INPUTDOMAIN%",
          "ttl":"3600"
       }
    ]
-}
+}
```
arnoldblinn: Where is the apex A record?  Don't they need this too?

myshopmatic.com.website.json
bushlack: Good catch - updated file name per suggestion.

myshopmatic.com.website.json
bushlack: I asked the provider about this.  They insisted they don't want one in the template for now.  They have 3 templates in GoDaddy's system that aren't in gitHub, and none of them have A records.

PR #47, opened 2018-07-24
Update Microsoft O365 template to use SPFM in v2

alinconstantin:


pawel-kow:
@alinconstantin @FlorentinStafie, are you sending the new variable already or shall we wait with publish till you do?

alinconstantin:
We're already sending both the v1-style SPFTXT variable (for providers that won't onboard immediately the v2 template), and the v2-style SPFRULES variable (for providers that will onboard the v2)

PR #48, opened 2018-07-25
Rename wordpress.com.g-suite.json to wordpress.com.wordpress-services…

pawel-kow:
….json

Adjustment to match the convention


pawel-kow:
Hi, last call for objections to this one.

arnoldblinn:
I have a more high level comment on this template.

I don't understand why Wordpress has their own template for setting up gsuite.  I created a gsuite-shared with the intent that all resellers of gsuite could use it.  This template allows name override; is synchronous only; and doesn't use signatures.  But it is pretty secure (no open IP addresses etc.).

Is there a reason why we need a one-off gsuite for wordpress?  I'll raise this in the slack too.

pawel-kow:
Obsolete after #79

PR #49, opened 2018-07-25
Update shopify.com.website.json

pawel-kow:
Fix Provider ID to the factual one


pawel-kow:
Hi, last call for objections to this one.

PR #54, opened 2018-08-16
Add template for Leadpages Custom Domain CNAME

timothy-hart:


leadpages.net-prod.custom-domain.json
```diff
@@ -0,0 +1,17 @@
+{
+   "providerId": "leadpages.net-prod",
```
pawel-kow: Hi, current convention for different environments is to use same providerId and different serviceId (see hubspot.com or gvng.com).

leadpages.net-prod.custom-domain.json
```diff
@@ -0,0 +1,17 @@
...
+   "serviceId": "custom-domain",
+   "serviceName": "Leadpages Custom Domain",
+   "version": 1,
+   "logoUrl": "https://my.leadpages.net/static/current/img/logos/logo_icon.svg",
+   "description": "Enables a domain to work with Leadpages",
+   "variableDescription": "DESTINATION: The subdomain for the domain.",
+   "destinationRequired": true,
+   "records": [{
+      "type": "CNAME",
+      "host": "%DESTINATION%",
```
pawel-kow: Please rather use "@" or "%host%" here, as this allows you to apply the same template many times without being replaced.
Please refer to this section of spec for details:
https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#65-variables-and-host

in your case, as host is always required (CNAME is not allowed on APEX), you'd also set "hostRequired": true

Generally you can base on this template which is very alike:
https://github.com/Domain-Connect/Templates/blob/master/app.unbounce.com.site.json

PR #56, opened 2018-08-23
Updated IP address for mapping a domain to about.me

NigelKibodeaux:


arnoldblinn:
This is a breaking change on the template. As such you should revise the version number. Version numbers were a more recent change.  This is an attribute "version": 2 on the template.

Also be mindful that changes to templates are not automatic with the DNS Providers.  You'll need to contact them to apply your update.

NigelKibodeaux:
Thanks @arnoldblinn. I got the version numbers updated.

PR #60, opened 2018-11-02
Add Login group

FlorentinStafie:
Add a new group named Login that allows the one to add email.contoso.com to point to specific URL for where the customer would get his email from.


jonroig:
Looks good. Thanks!

PR #74, opened 2019-05-22
Bingwebmastertool domainconnect template deployment with returnurl

Jainabgitdev:


bing.com.BingWebMasterDPOOV1.json
```diff
@@ -0,0 +1,22 @@
+{
+  "providerId": "bing.com",
+  "providerName": "Microsoft Bing",
+  "serviceId": "BingWebMasterDPOOV2",
```
pawel-kow: Is it new template, or a new version of BingWebMasterDPOO ?
The difference is the added syncRedirectDomain, which allows you to use with or without it, so for me it seems backward compatible, so no need to create a new template.
Please also keep the file name in sync with the structure {providerId}.{serviceId}.json

bing.com.BingWebMasterDPOOV1.json
```diff
@@ -0,0 +1,22 @@
...
+  "serviceName": "Domain proof of ownership with Bing Webmaster services",
+  "version": 1,
+  "logoUrl": "https://domainconnect.org/wp-content/uploads/<<bing logo>> link",
+  "description": "Verifies domain ownership for further configuration of Microsoft Bing Services",
+  "variableDescription": "Guid of webmaster for verification of domain ownership with Bing webmaster enables quick and easy SEO on Bing.",
+  "syncBlock": false,
+  "shared": false,
+  "warnPhishing": false,
+  "syncRedirectDomain": "bing.com/webmaster/configure/verify/ownership",
```
pawel-kow: Here it's just a domain name, so pls correct to "syncRedirectDomain": "bing.com"

bing.com.BingWebMasterDPOOV1.json
```diff
@@ -0,0 +1,22 @@
...
+  "records": [
+      {  
+         "type":"CNAME",
+         "host":"%WEBMASTERGUID%",
+         "pointsTo":"verify.bing.com",
+         "ttl":"604800"
```
pawel-kow: Sure about such huge TTL? What is the motivation here?
I think 7 days may not be supported by some providers. We for example don't support longer than 2 days.

bing.com.BingWebMasterDPOOV1.json
```diff
@@ -0,0 +1,22 @@
+{
+  "providerId": "bing.com",
+  "providerName": "Microsoft Bing",
+  "serviceId": "BingWebMasterDPOOV2",
+  "serviceName": "Domain proof of ownership with Bing Webmaster services",
+  "version": 1,
```
pawel-kow: If you'll change the serviceId back to BingWebMasterDPOO (see comment above) please bump the version here.

bing.com.BingWebMasterDPOOV1.json
Jainabgitdev: Changed to 2 days as for now, supporting only 1&1 DNS provider

bing.com.BingWebMasterDPOOV1.json
Jainabgitdev: Sure

bing.com.BingWebMasterDPOO.json
```diff
@@ -0,0 +1,22 @@
+{
+  "providerId": "bing.com",
+  "providerName": "Microsoft Bing",
+  "serviceId": "BingWebMasterDPOO",
+  "serviceName": "Domain proof of ownership with Bing Webmaster services",
+  "version": 1.2,
```
pawel-kow: Pls correct version to 2, as it is defined to be an integer.

bing.com.BingWebMasterDPOO.json
```diff
@@ -0,0 +1,22 @@
...
+  "logoUrl": "https://domainconnect.org/wp-content/uploads/<<bing logo>> link",
```
pawel-kow: Mind adding a valid URL.

bing.com.BingWebMasterDPOO.json
```diff
@@ -0,0 +1,22 @@
+{
+  "providerId": "bing.com",
+  "providerName": "Microsoft Bing",
+  "serviceId": "BingWebMasterDPOO",
```
arnoldblinn: I don't like the mixed case serviceId name.  Same with the file name.  Can we use bing.com.dpoo.json?  This is what the other microsoft template did for o365.

bing.com.BingWebMasterDPOO.json
arnoldblinn: You can also omit it.  The field isn't required.

bing.com.BingWebMasterDPOO.json
```diff
@@ -0,0 +1,22 @@
...
+  "description": "Verifies domain ownership for further configuration of Microsoft Bing Services",
+  "variableDescription": "Guid of webmaster for verification of domain ownership with Bing webmaster enables quick and easy SEO on Bing.",
+  "syncBlock": false,
+  "shared": false,
+  "warnPhishing": false,
+  "syncRedirectDomain": "www.bing.com",
+  "records": [
+      {  
+         "type":"CNAME",
+         "host":"%WEBMASTERGUID%",
```
arnoldblinn: In general we aren't super excited about variables in the host field.  However, the spec does make an exception for a cname used for validation.  So I'm OK with it.

bing.com.BingWebMasterDPOO.json
```diff
@@ -0,0 +1,22 @@
...
+  "serviceId": "BingWebMasterDPOO",
+  "serviceName": "Domain proof of ownership with Bing Webmaster services",
+  "version": 2,
+  "logoUrl": "https://domainconnect.org/wp-content/uploads/<<bing logo>> link",
+  "description": "Verifies domain ownership for further configuration of Microsoft Bing Services",
+  "variableDescription": "Guid of webmaster for verification of domain ownership with Bing webmaster enables quick and easy SEO on Bing.",
+  "syncBlock": false,
+  "shared": false,
+  "warnPhishing": false,
+  "syncRedirectDomain": "www.bing.com",
```
arnoldblinn: Are you sure you want syncRedirectDomain to contain www?  This means any redirect url passed in would need to have a domain of the form www.bing.com.  bing.com (or foo.bing.com) wouldn't work.

Most templates ommit the www.

bing.com.BingWebMasterDPOO.json
```diff
@@ -0,0 +1,22 @@
+{
+  "providerId": "bing.com",
```
arnoldblinn: This template as defined does not require signatures.  I'm not sure what the fishing thread here is; could I send a phishing mail out with my GUID, get the domain owner to click my link and validate me, and then have access to domain level functionality?  Read section 4.2.3 of the spec for more information.  I'm guessing you'll want to either sign your requests, or at least turn on warnPhising.....

bing.com.BingWebMasterDPOO.json
pawel-kow: yes, their redirect url is with www

bing.com.BingWebMasterDPOO.json
pawel-kow: @arnoldblinn with the current language in the spec this field is case-insensitive anyway, so shall we bother anymore?

arnoldblinn: Indeed, the values pass in the query string for the provider/service names are case insensitive. Largely because people started doing it.  But for template names, they are case sensitive.  All other template names are lower case, making the mapping of the queyr string values to a file name easy.  I do think we should make this lower case.

bing.com.BingWebMasterDPOO.json
arnoldblinn: But why would they want to limit this?  By putting in www.bing.com, the redirect can only be to www.bing.com.  If you put in bing.com, you could redirect to www.bing.com or foo.bing.com. Doesn't matter much at the end of the day I guess.

pawel-kow: AFAIK we don't have such language in the spec to match also subdomains for syncRedirectDomain.
Assuming that more specific is more secure for redirect_urls, it's reasonable to keep the exact match requirement for syncRedirectDomain. In case more hosts are needed, it is comma-separated list for this purpose.

bing.com.BingWebMasterDPOO.json
Jainabgitdev: Agree. Resolved

bing.com.BingWebMasterDPOO.json
Jainabgitdev: We need this to be variable, we notify users the actual value of this field before making DNS changes. So, this should be okay to keep host as variable.

bing.com.BingWebMasterDPOO.json
Jainabgitdev: Done.

bing.com.bingwebmasterdpoo.json
```diff
@@ -0,0 +1,21 @@
+{
+  "providerId": "bing.com",
+  "providerName": "Microsoft Bing",
+  "serviceId": "bingwebmasterdpoo",
+  "serviceName": "Domain proof of ownership with Bing Webmaster services",
```
arnoldblinn: Wonder if you want something shorter here.  Note: Right now this string is displayed to the user, and there is not a mechanism in the specification for localization.  So you probably want something simpler....

pawel-kow: 👍 for this one. A shorter name and something that would fit better even in localized texts would be a way better.

Jainabgitdev: Agree, changed.

bing.com.bingwebmasterdpoo.json
```diff
@@ -1,8 +1,8 @@
 {
   "providerId": "bing.com",
   "providerName": "Microsoft Bing",
-  "serviceId": "bingwebmasterdpoo",
-  "serviceName": "Domain proof of ownership with Bing Webmaster services",
+  "serviceId": "BingWebMasterDPOO",
+  "serviceName": "domain ownership with Bing Webmaster services",
```
pawel-kow: How about still shorter: "Bing Webmaster services" ?

Jainabgitdev: Done

PR #75, opened 2019-06-14
Make variables in O365 templates case sensitive/uppercase, shared templates

alinconstantin:
Make variables in O365 templates case sensitive/uppercase, change templates to shared


arnoldblinn:
Do you want to rev version for this one?  I could go either way.

alinconstantin:
> Do you want to rev version for this one?  I could go either way.

No, I don't think it's necessary this time.

pawel-kow:
+1 LGTM

PR #91, opened 2019-10-10
Adobe Portfolio

TracyFogel:
Initial version of templates for Adobe Portfolio. We use both a production and a stage environment.

We plan to make some additions later, including adding a `logoUrl` once we have that hosted somewhere.

### DIFF
--- a/portfolio.adobe.com.prod.json
+++ b/portfolio.adobe.com.prod.json
@@ -2,7 +2,7 @@
   "providerId":"portfolio.adobe.com",
   "providerName":"Adobe Portfolio",
   "serviceId":"prod",
-  "serviceName":"Adobe Portfolio Websites",
+  "serviceName":"Portfolio Site",
   "version":1,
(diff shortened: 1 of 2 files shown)


portfolio.adobe.com.prod.json
```diff
@@ -0,0 +1,40 @@
+{
+  "providerId":"portfolio.adobe.com",
+  "providerName":"Adobe Portfolio",
+  "serviceId":"prod",
+  "serviceName":"Adobe Portfolio Websites",
```
pawel-kow: A very minor comment, nothing blocking, just to consider if the redundancy of "Adobe Portfolio" between `providerName` and `serviceId` is needed.

This is how it would look on IONOS consent screen:
image

TracyFogel: Thanks for pointing this out. We'll push an updated serviceName shortly.

TracyFogel: Updated. Thanks!

PR #92, opened 2019-10-18
Portfolio: TXT record

TracyFogel:
Adding a TXT record, plus setting warnPhishing = true for additional security with the introduction of a variable in our data.


pawel-kow:
Hi @TracyFogel , is the solution with the TXT record something your systems are using already, or a new feature you are developing?
If the content of the record is something you can influence, I would propose to add some specific part, which won't be delivered by a variable, in order to avoid the need for `warnPhishing` setting.
Typically this can be achieved by using key=value notation. Alternatively putting the record on underscore subdomain may be a good idea.

If not possible, pity, we stay with `warnPhishing` and template as proposed.

arnoldblinn:
Agree largely with Pawel on this one.  Having a TXT record at the root of the zone is something that can cause problems down the line with conflicts.  Other services may define TXT records at the root of the zone, and having no unique value indicating it is you can be problematic.

I've seen people add something to the record content such as 'adobeval=%uuid%', or name the record something unique (e.g. the name of the host is _adobveval, not @).

A validation record rarely warrants a warning from phising though.  It is typically placed in a zone to control who owns the domain when two users both claim to own.

I suppose you could run into a situation where one user owns the domain, sets up the site, and another user comes in and CLAIMS they own the domain and fishes the actual owner to update the unique id to the bad actor.  Then they could take over the hosting.  

On that front it would be waranted.  

We always recomend digital signatures over the warning though.

TracyFogel:
Thanks @pawel-kow and @arnoldblinn for the excellent suggestions. We are able to support a key=%value% structure for the TXT record data so that part of it is non-variable. That's been updated, as well as a host subdomain. (And a logo!)

PR #111, opened 2020-03-17
Add first version of the myPaymail Domain Connect template

polmyeste:
We would like to add a template for our service so that a DNS provider can host it.

### DIFF
--- a/trustchair.com.mypaymail.json
+++ b/trustchair.com.mypaymail.json
@@ -1,0 +1,31 @@
+{
+  "providerId": "trustchair.com",
+  "providerName": "Trustchair Ltd",
+  "serviceId": "myPaymail",
+  "serviceName": "myPaymail",
+  "version": 1,
+  "logoUrl": "https://mypaymail.co/images/DomainConnectLogo.png",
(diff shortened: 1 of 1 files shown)


pawel-kow:
Hi @polmyeste 
Generally all looks pretty good, just 2 clarifying questions:
1. why you think it's needed to set `warnPhishing: true`. The template looks like being quite specific to your service, so not seeing a vector to misuse it ti connect to other malicious service
2. Do you consider using `redirect_uri` in the sync flow? Consider `syncRedirectDomain` must be set if you do.

polmyeste:
Hi Pawel,

Thanks for reviewing, my replies below:

1. I have set the flag because I read be following from the spec: _If the template require variables, and OAuth and signing isn’t available, the flag warnPhishing must be set to true in the template._ The target field of the SRV record is quite important for our service as its value is trusted when clients read it. The customer will double-check the target value when they are enabling DNSSEC in the first release. After the beta release of the product, we may look at signing the template.

2. I have considered it, but at the moment I am detecting if the user closes a popup window.

PR #117, opened 2020-04-16
Create a new template to connect A record

moxlive:
This PR is to create a new template to add 2 A records for one Apex domain. This template will be used in both QA and PROD environments.

### DIFF
--- a/hubspot.com.hubspot-template-3.json
+++ b/hubspot.com.hubspot-template-3.json
@@ -1,0 +1,27 @@
+{
+    "providerId": "hubspot.com",
+    "providerName": "HubSpot",
+    "serviceId": "hubspot-template-3",
+    "serviceName": "HubSpot CMS Site",
+    "version": 1,
+    "logoUrl": "https://domainconnect.org/wp-content/uploads/2017/02/HubSpot-768x223.png",
+    "description": "Enables an apex domain to work with the HubSpot CMS",
+    "variableDescription": "IPs: static IPs owned by hubspot CMS",
+    "hostRequired": true,
+    "records": [{  
+         "type":"A",
+         "host":"@",
+         "pointsTo":"%IP1%",
+         "ttl":600
+    },{  
+         "type":"A",
+         "host":"@",
+         "pointsTo":"%IP2%",
+         "ttl":600
+    },{  
+         "type":"A",
+         "host":"@",
+         "pointsTo":"%IP3%",
+         "ttl":600
+    }]
+}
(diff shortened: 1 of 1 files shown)


hubspot.com.hubspot-template-3.json
```diff
@@ -0,0 +1,22 @@
+{
+    "providerId": "hubspot.com",
+    "providerName": "HubSpot",
+    "serviceId": "hubspot-template-3",
+    "serviceName": "HubSpot CMS Site",
+    "version": 1,
+    "logoUrl": "https://domainconnect.org/wp-content/uploads/2017/02/HubSpot-768x223.png",
+    "description": "Enables an apex domain to work with the HubSpot CMS",
+    "variableDescription": "IPs: HubSpot CMS IPs",
+    "hostRequired": true,
```
arnoldblinn, line 10: Why is hostRequired set to true?  Usually this is required when the template doesn't work with an apex domain (e.g. you have a cname record using the fqdn, and the template would only work with a host/sub-domain).

moxlive: Good catch, I will remove it

moxlive:
> Disable Synchronous
One option is to disable the synchronous flow and use asynchronous OAuth. This can be controlled with the syncBlock value from the template. However, as will be seen below OAuth has a higher implementation burden and requires onboarding between each Service and DNS Provider.

Hubspot uses OAuth flow only, I can disable syncBlock then

moxlive:
warnPhishing seems to be a good feature to enable anyway
>(optional) When present, this tells the DNS Provider that the template may contain variables susceptible to phishing attacks and the provider is unable to digitally sign the requests. When set the DNS Provider should display warnings to the user. The default value for this is false.
@arnoldblinn do you mind sharing a little more how the phishing is detected and what the warning looks like? Does GoDaddy maintain a list of phishing sites?

arnoldblinn:
@moxlive Couple of things.  If you are oauth only, then set syncBlock to true.  

If a template is synchronous, a hacker could format a properly formatted URL to apply your template and send it in (say) a phishing email.  They would/could replace the IP address in the URL with their own.

One mitigation to this is digitally signing the request (syncPubKeyDomain).  If this isn't possible, the other mitigation is to set warnPhishing.  When this flag is set, it tells the DNS Provider (e.g. GoDaddy) to render an extra warning on the confirmation screen.

For example: Please verify the link was from a reputable source.

Phising sites aren't maintained.  It simply tells the DNS Provider to render the extra warning.

Since you are async only, set syncBlock to true.

moxlive:
@arnoldblinn  thanks for reviewing this PR.  Do you mind taking another look when getting a chance?

moxlive:
@arnoldblinn  thanks for merging it. Do you have an estimation on when will this new template be ready for testing?

arnoldblinn:
@moxlive: I resigned from GoDaddy several weeks ago.  You'll need to ping them directly.  I would suggest sending mail to domainconnect@godaddy.com, or hitting them up in the slack.

moxlive:
@arnoldblinn  thank you for the advice. I wish you all the best

PR #119, opened 2020-04-25
Num templates

twalmsley:
Add the `syncPubKeyDomain` because we will be using the synchronous flow with digital signatures.

### DIFF
--- a/numserver.com.custodian-record.json
+++ b/numserver.com.custodian-record.json
@@ -9,6 +9,7 @@
    "logoUrl":"",
    "description":"Configure a NUM custodian record for your domain name to prove that you are an administrator of the domain, for more info see: https://num.uk/modules/4",
    "variableDescription":"The user_hash variable is a SHA-1 hash of an identifier, e.g. email, domain or telephone number",   
+   "syncPubKeyDomain":"numserver.com",
    "records":[
       {
          "type":"TXT",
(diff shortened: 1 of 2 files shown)


numserver.com.custodian-record.json
```diff
@@ -3,12 +3,14 @@
    "providerName":"NUM Server",
    "serviceId":"custodian-record",
    "serviceName":"Custodian Record",
+   "logoUrl": "http://numserver.com/images/logo-black.png",
    "version": 1,
    "shared": true,
    "sharedProviderName": true,
    "logoUrl":"",
```
pawel-kow, line 10: Noticed only now, `logoUrl` is doubled here. Not sure how it passes schema check... but still pls fix it ;)

numserver.com.delegate-num-zone.json
```diff
@@ -3,12 +3,14 @@
    "providerName":"NUM Server",
    "serviceId":"delegate-num-zone",
    "serviceName":"NUM Zone Delegation",
+   "logoUrl": "http://numserver.com/images/logo-black.png",
    "version": 1,
    "shared": true,
    "sharedProviderName": true,
    "logoUrl":"",
```
pawel-kow, line 10: Noticed only now, `logoUrl` is doubled here. Not sure how it passes schema check... but still pls fix it ;)

PR #128, opened 2020-05-20
Titan Email w/ cPanel GD templates

mr-brik:
Titan Email replaces FlockMail; see secureserver.net.cpanel_flockmail.json + godaddy.com.cpanel_flockmail.json.

### DIFF
--- a/godaddy.com.cpanel_titanemail.json
+++ b/godaddy.com.cpanel_titanemail.json
@@ -1,0 +1,39 @@
+{
+  "providerId": "godaddy.com",
+  "providerName": "GoDaddy",
+  "serviceId": "cpanel_titanemail",
+  "serviceName": "cPanel TitanEmail",
+  "version": 1,
+  "logoUrl": "https://www.domainconnect.org/wp-content/uploads/2020/02/godaddy_logo_new.png",
+  "description": "Enables a domain to work with TitanEmail",
+  "variableDescription": "",
+  "syncPubKeyDomain": "domain-attach.api.godaddy.com",
+  "records": [
+    {
+      "type": "MX",
+      "host": "@",
+      "pointsTo": "mx1.titan.email",
+      "priority": 10,
+      "ttl": 3600
+    },
+    {
+      "type": "MX",
+      "host": "@",
+      "pointsTo": "mx2.titan.email",
+      "priority": 20,
+      "ttl": 3600
+    },
+    {
+      "type": "TXT",
+      "host": "@",
+      "data": "v=spf1 include:spf.titan.email ~all",
+      "ttl": 3600
+    },
+    {
+      "type": "TXT",
+      "host": "_dmarc",
+      "data": "v=DMARC1;p=none;sp=none;adkim=r;aspf=r;pct=100",
+      "ttl": 3600
+    }
+  ]
+}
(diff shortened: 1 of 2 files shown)


godaddy.com.cpanel_titanemail.json
```diff
@@ -0,0 +1,39 @@
...
+      "pointsTo": "mx2.titan.email",
+      "priority": 20,
+      "ttl": 3600
+    },
+    {
+      "type": "TXT",
+      "host": "@",
+      "data": "v=spf1 include:spf.titan.email ~all",
+      "ttl": 3600
+    },
```
arnoldblinn, line 31: It is preferred that Domain Connect templates use the SPFM record, not a TXT record for SPF.  Otherwise conflicting SPF records in templates would overwrite each other.

godaddy.com.cpanel_titanemail.json
```diff
@@ -0,0 +1,39 @@
...
+    {
+      "type": "TXT",
+      "host": "_dmarc",
+      "data": "v=DMARC1;p=none;sp=none;adkim=r;aspf=r;pct=100",
+      "ttl": 3600
```
arnoldblinn, line 36: We have not seen any other templates set a DMARC record; largely because of the same conflicts that would likely occur with SPF (there can only be one DMARC record). Most other templates have removed it.  You can leave it, but it will begin to slippery slope us toward DMARC merging like SPF>

secureserver.net.cpanel_titanemail.json
```diff
@@ -0,0 +1,39 @@
...
+      "type": "MX",
+      "host": "@",
+      "pointsTo": "mx2.titan.email",
+      "priority": 20,
+      "ttl": 3600
+    },
+    {
+      "type": "TXT",
+      "host": "@",
+      "data": "v=spf1 include:spf.titan.email ~all",
```
arnoldblinn, line 29: See above.

secureserver.net.cpanel_titanemail.json
```diff
@@ -0,0 +1,39 @@
...
+      "ttl": 3600
+    },
+    {
+      "type": "TXT",
+      "host": "_dmarc",
+      "data": "v=DMARC1;p=none;sp=none;adkim=r;aspf=r;pct=100",
+      "ttl": 3600
+    }
```
arnoldblinn, line 37: See above

godaddy.com.cpanel_titanemail.json
darren-locke: Hi Arnold, these are pretty much mirror copies of existing templates (https://github.com/Domain-Connect/Templates/blob/master/secureserver.net.cpanel_flockmail.json) which have been updated to reflect the change in branding from Flock to Titan email. The existing ones are using TXT for SPF and also have the DMARC record

arnoldblinn: You are right; these use dmarc.  And I see one other template from plesk using it.  If critical, leave it in.

But nobody is using SPF.  Convert the new ones to SPFM.  Will behave better.

godaddy.com.cpanel_titanemail.json
mr-brik: It's similar but at least you don't have the same issue with compound values; indeed you might have the same issue for setting a _domainconnect record! I might be mistaken but I think the Domain Connect spec had some provision for such records.

PR #132, opened 2020-07-22
Update hubspot.com.hubspot-template-2.json

jgrace03:
Update the hubspot-template-2 to add variables for the CNAME hosts and CNAME value suffixes

@moxlive 

Our goal is to support the following patterns in a single template for CNAME hosts. 

Pattern 1 PROD:
hs1-\<id\>
hs2-\<id\>

Pattern 2 PROD:
hs1
hs2

Pattern 1 QA:
hs-qa1-\<id\>
hs-qa2-\<id\>

Pattern 2 QA:
hs-qa1
hs-qa2

### DIFF
--- a/hubspot.com.hubspot-template-2.json
+++ b/hubspot.com.hubspot-template-2.json
@@ -10,14 +10,14 @@
   "records": [
     {
       "type": "CNAME",
-      "host": "hs1",
-      "pointsTo": "%cname0targetprefix%.dkim.hubspotemail.net",
+      "host": "%cname0host%",
+      "pointsTo": "%cname0targetprefix%.dkim.hubspotemail%envsuffix%",
       "ttl": 3600
      },
      {
        "type": "CNAME",
-       "host": "hs2",
-       "pointsTo": "%cname1targetprefix%.dkim.hubspotemail.net",
+       "host": "%cname1host%",
+       "pointsTo": "%cname1targetprefix%.dkim.hubspotemail%envsuffix%",
        "ttl": 3600
      },
      {
--- a/hubspot.com.hubspot-template-2.json
+++ b/hubspot.com.hubspot-template-2.json
@@ -10,13 +10,13 @@
     {
       "type": "CNAME",
       "host": "@",
-      "pointsTo": "%DESTINATION0%",
+      "pointsTo": "%cnamedestination0%",
       "ttl": 3600
      },
      {
        "type": "CNAME",
        "host": "@",
-       "pointsTo": "%DESTINATION1%",
+       "pointsTo": "%cnamedestination1%",
        "ttl": 3600
      },
      {
--- a/hubspot.com.hubspot-template-2.json
+++ b/hubspot.com.hubspot-template-2.json
@@ -10,13 +10,13 @@
     {
       "type": "CNAME",
       "host": "@",
-      "pointsTo": "%cnamedestination0%",
+      "pointsTo": "%destination0%",
       "ttl": 3600
      },
      {
        "type": "CNAME",
        "host": "@",
-       "pointsTo": "%cnamedestination1%",
+       "pointsTo": "%destination1%",
        "ttl": 3600
      },
      {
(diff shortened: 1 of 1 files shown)


jgrace03:
@arnoldblinn thanks for the review. I added more information about what our requirement is for the CNAME hosts in the description above. I'd appreciate any recommendations you may have. 

We run into this case from the spec:
"Note: There are some templates that utilize CNAME or TXT records with host values containing some form of user identification for validation of domain ownership, and these are often passed in variables.

To support this use case, variables are allowed for the host name. But only in this limited circumstance."

arnoldblinn:
It this the case here?  Usually this manifests as a verification record of some sort; where the host name is a user specific guid or something. An example of this is in this template:

https://github.com/Domain-Connect/Templates/blob/885d164ab5dda7e1f2e590051f8d952f8228e860/bing.com.bingwebmasterdpoo.json

Even many service providers are moving away from verification being in the host name; largely for reasons similar to this.  Instead, they put the random value in the data itself.  Microsoft O365 and GSuite moved this direction.

So my question is this: is the intent of the variable host to put a random user-specific value in there? Or is it to put a true hostname in there that the user would actually type in the browser?

jgrace03:
Would something like `host: hs%env%1%id%` work if we were to exclude the env and id variables when they were not needed?

jgrace03:
This is the case. The id in `hs-qa1-\<id\>` is user specific and used for verifying identity. 

Another challenge, for the same root domain, we'd like to allow, for example,  `hs-qa1-1` and `hs-qa1-2` to be added the same record at different times. If we use the same template, will one attempt to replace the other?

arnoldblinn:
Exclusion of variables would cause an error in most implementations.

If the record is in fact used for verification with some sort of GUID, we can look at approving this.  Other providers do this. It is the exception to the rule.

Your hs-qa-1 and hs-qa-2 questions are spot on.  For some providers, applying one would in fact wipe out the other.  This would be true for GoDaddy or for 1&1.  For other providers, it would not.

THis might be better to take offline and have a conversation on; there are other providers that have solved this problem with TXT records, including some features to enable multiple applications.  But this is way in the weeds of the spec and back and forth in github isn't going to be productive.

Happy to jump on a call.  Are you on our slack channel? Hit me up there.

jgrace03:
I am! I will consult with my team to iron out some of these detail a bit more and hit you up when I'm ready to chat in more detail. Thank you!

hubspot.com.hubspot-template-2.json
```diff
@@ -6,18 +6,17 @@
   "version": 1,
   "logoUrl":"https://domainconnect.org/wp-content/uploads/2017/02/HubSpot-768x223.png",
   "description":"Enables a domain to work with HubSpot",
-  "variableDescription":"v1 is the customer id.",
   "records": [
```
arnoldblinn, line 9: Your travis build is failing; I think because you need to specify "hostRequired" : true in the template.  If you set a CNAME = @, the host must be passed in with the template.  This is because you can't CNAME a root domain.

I think this is why the build is failing

pawel-kow: Not only that. It is also not allowed (by DNS RFC) to have 2 CNAME RRs on the same host.

arnoldblinn: Right. I missed this.  From our conversation, couple of things.  First, not sure why you needed two records in the template.  If I recall, you want multiple teams calling and applying the template.  Not sure why they can't use the same variable name in the template.  This being said, based on Pawel's catch, you are also trying to have multiple authorized users of your domain.  You probably want to switch to TXT records for this.  It is similar to what Google is doing with TXT records.

jgrace03:
Thanks @arnoldblinn and @pawel-kow for the feedback. I am going to close this PR as we are going to go in a different direction here.

jgrace03:
Thanks @arnoldblinn and @pawel-kow for the feedback. I am going to close this PR as we are going to go in a different direction here.

jgrace03:
Thanks @arnoldblinn and @pawel-kow for the feedback. I am going to close this PR as we are going to go in a different direction here.

pawel-kow:
Based on the last comment from @jgrace03 I close it.
Please open a new PR once a new approach is taken.

PR #133, opened 2020-07-27
add templates for realgeeks website/crm service

slyboots:
Templates for Real Geeks web platform

### DIFF
--- a/realgeeks.com.website_and_crm.json
+++ b/realgeeks.com.website_and_crm.json
@@ -1,0 +1,70 @@
+{
+  "providerId": "realgeeks.com",
+  "providerName": "Real Geeks",
+  "serviceId": "website_and_crm",
+  "serviceName": "Real Geeks Real Estate Websites",
+  "version": 1,
+  "logoUrl": "https://www.realgeeks.com/static/img/logo_2x.png",
+  "description": "Enables a domain to be used for a Real Geeks website and CRM",
+  "variableDescription": "Self explanatory in the variable names.",
+  "syncPubKeyDomain": "domainconnect.realgeeks.com",
+  "records": [
+    {
+      "groupId": "Website",
+      "type": "APEXCNAME",
+      "host": "@",
+      "pointsTo": "%prodhost%",
+      "ttl": 600
+    },
+    {
+      "groupId": "Website",
+      "type": "CNAME",
+      "host": "www",
+      "pointsTo": "ssl.realgeeks.com",
+      "ttl": 600
+    },
+    {
+      "groupId": "CRM",
+      "type": "MX",
+      "host": "rg-mail.www",
+      "pointsTo": "mxa.mailgun.org",
+      "priority": 10,
+      "ttl": 300
+    },
+    {
+      "groupId": "CRM",
+      "type": "MX",
+      "host": "rg-mail.www",
+      "pointsTo": "mxb.mailgun.org",
+      "priority": 10,
+      "ttl": 300
+    },
+    {
+      "groupId": "CRM",
+      "type": "SPFM",
+      "host": "rg-mail.www",
+      "spfRules": "include:mailgun.org"
+    },
+    {
+      "groupId": "CRM",
+      "type": "CNAME",
+      "host": "email.rg-mail.www",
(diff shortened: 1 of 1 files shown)


realgeeks.com.website_and_crm.json
```diff
@@ -0,0 +1,69 @@
...
+      "pointsTo": "mxb.mailgun.org",
+      "priority": 10,
+      "ttl": 300
+    },
+    {
+      "groupId": "CRM",
+      "type": "SPFM",
+      "host": "rg-mail.www",
+      "spfRules": "v=spf1 include:mailgun.org ~all"
+    },
```
arnoldblinn, line 46: With SPFM, you don't want the v=spf1 or ~all. See 6.10.3 in the spec

realgeeks.com.website_and_crm.json
```diff
@@ -0,0 +1,69 @@
...
+  "description": "Enables a domain to be used for a Real Geeks website and CRM",
+  "variableDescription": "Self explanatory in the variable names.",
+  "records": [
+    {
+      "groupId": "Website",
+      "type": "APEXCNAME",
+      "host": "@",
+      "pointsTo": "%PRODHOST%",
+      "ttl": 600
+    },
```
arnoldblinn, line 17: APEXCNAME will severely limit you on DNS Providers you can work with; but this is your perogative.

Also recommend that you make variable names lower case....experience tells me this is a bit more friendly.

You also are going to want to digitally sign this template.  A synchronous template with a variable in pointsTo is susceptible to fishing attacks.  See section 4.2.3 of the spec

realgeeks.com.website_and_crm.json
slyboots: Ahh, thank you!

realgeeks.com.website_and_crm.json
slyboots: The APEXCNAME is actually one of the requirements we had for using Domain Connect. Luckily GoDaddy supports it, and they're currently the only ones we're wanting to work with.

The caps lock variables felt bad to type honestly.

This is meant to be a synchronous template, so it definitely needs to be signed. Thanks for the input!

slyboots: Added the syncPubKeyDomain key to the template.

arnoldblinn:
We had to update the build (which is a schema check) to make your pull request pass our Travis CI build.  The schema check didn't like the APEXCNAME (which you are really the first company to depend on....like I said use of this fields severely limits which DNS Providers you can work with). Please merge master back and then I can merge your pull request.

slyboots:
I've merged master back in. It seems like the build is still failing though...
I didnt realize what we were wanting was so far outside of the norm since it was mentioned in the specification.
I really do appreciate your help despite the extra work Im causing 😅

arnoldblinn:
https://travis-ci.com/github/Domain-Connect/Templates/builds/177478977

The error is that your serviceId doesn't match the name of the file.  This is required.  Either change your service id, or change the file name.

pawel-kow:
> I've merged master back in. It seems like the build is still failing though...

This  time it's file naming issue

`realgeeks.com.website_and_crm.json not matching the pattern. Expected realgeeks.com.realgeeks.json`

pawel-kow:
Btw... well APEXCNAME is something barely anyone supports. For a reason it's in extension part odd the spec.  Main problem is it does not work with pre signed zones with DNSSEC.
I understand you want to work with GD but in fact investing in Domain Connect it's a pity and a bit short sighted AFAIK as other providers would work out of the box.

slyboots:
Ahh, I didnt realize I could actually see the build results until just now. I should have clicked around a bit more. Thank you both.

>I understand you want to work with GD but in fact investing in Domain Connect it's a pity and a bit short sighted AFAIK as other providers would work out of the box.
I dont disagree with you, but ~80% of our client base uses godaddy. We cant really ask them to switch to a different registrar. Changing our infrastructure so that we dont require an APEXCNAME is easier said than done though. So If we have an opportunity to improve our support for the majority of our clients with minimal changes right now it seems reasonable to take advantage of it and then focus on further improving how we use Domain Connect.

PR #136, opened 2020-08-04
Add Website Toolbox templates

webtoolbox:


websitetoolbox.com.apexforum.json
```diff
@@ -0,0 +1,36 @@
+{
+  "providerId":"WebsiteToolbox.com",
```
arnoldblinn, line 2: Please make lower case.

websitetoolbox.com.apexforum.json
```diff
@@ -0,0 +1,36 @@
...
+      "target":"https://www.%fqdn%"
+    },
+    {
+      "type":"CNAME",
+      "host":"www",
+      "pointsTo":"custom.websitetoolbox.com",
+      "ttl":3600
+    },
+    {
+      "type": "TXT",
```
arnoldblinn, line 24: For an SPF record you should use the SPFM type.  This will allow multiple templates that try to set SPF records to co-exist.

websitetoolbox.com.apexforum.json
```diff
@@ -0,0 +1,36 @@
...
+  "serviceId":"apexforum",
+  "serviceName":"Website Toolbox Forum",
+  "version": 1,
+  "syncPubKeyDomain" : "websitetoolbox.com",
+  "logoUrl":"https://cdn.websitetoolbox.com/images/logos/wt_logo_transparent.png",
+  "description":"Enables the root of a domain to work with Website Toolbox.",
+  "variableDescription":"",
+  "records":[
+    {
+      "type":"REDIR301",
```
arnoldblinn, line 13: Not all DNS Providers support the REDIR command; so be warned that this template may not work well with all the DNS Providers.

websitetoolbox.com.forum.json
```diff
@@ -0,0 +1,31 @@
...
+  "serviceName":"Website Toolbox Forum",
+  "version": 1,
+  "syncPubKeyDomain" : "websitetoolbox.com",
+  "logoUrl":"https://cdn.websitetoolbox.com/images/logos/wt_logo_transparent.png",
+  "description":"Enables a subdomain to work with Website Toolbox.",
+  "variableDescription":"v1 is the subdomain that wants to be used. ie: 'forum' for forum.mydomain.com.",
+  "records":[
+    {
+      "type":"CNAME",
+      "host":"%v1%",
```
arnoldblinn, line 14: Using variables as a host name is not recommended.  This is outlined in the spec in section 6.9.1.

websitetoolbox.com.forum.json
```diff
@@ -0,0 +1,31 @@
+{
+  "providerId":"WebsiteToolbox.com",
```
arnoldblinn: Lower case please.

websitetoolbox.com.forum.json
```diff
@@ -0,0 +1,31 @@
...
+  "variableDescription":"v1 is the subdomain that wants to be used. ie: 'forum' for forum.mydomain.com.",
+  "records":[
+    {
+      "type":"CNAME",
+      "host":"%v1%",
+      "pointsTo":"custom.websitetoolbox.com",
+      "ttl":3600
+    },
+    {
+      "type": "TXT",
```
arnoldblinn, line 19: SPFM please.

websitetoolbox.com.forum.json
webtoolbox: Thank you for all your feedback!

We don't know the host ahead of time. The client can choose any host. For example, forums.mydomain.com, forum.mydomain.com, community.mydomain.com, etc.

They would only have one domain set up at a time, so if a provider needs to remove the old template before setting the new one (as the documentation mentions), that would be ok.

arnoldblinn: Hmmm.  This is an interesting use case.  

So you are applying the template to mydomain.com.  You need an SPF record modified at mydomain.com.  You want websitetoolbox._domainkey.mydomain.com to be set.  But you also want an arbitrary CNAME as a sub-domain set to mydomain.com. 

The template would be applied to mydomain.com.  So the old value would in fact be removed. The challenge will be if the DNS Providers can do conflict detection.  Some of the DNS Providers don't like arbitrary variables in host names because it makes detecting conflicts and ifnorming the user harder.  For example, if there was already an entry at form.mydomain.com and you passed in a variable of forum.

This is really problematic on async templates.  For synchronous, you might get away with this.  But it is a bit unorthodox compared to other templates I've seen.  Would be curious what @pawel-kow thinks about this.

arnoldblinn: Another option; albeit not without drawbacks; is to use async.  You could get an async request with permission to apply two templates and on two domains/sub-domains.  First template you would apply to the root, second one to the sub-domain.

The downside is you need to go async to do this, which also limits the DNS Providers you could work with.  Probably less restrictive than the redir command you are relying on though.

webtoolbox: Yes, that's correct.

We would be using synchronous flow.

If there's a better way to do this, we're all ears. But it's super important for us to allow the client to choose the subdomain they want to use for their forum. Some, for example, speak Spanish and might event want "foro" instead of "forum". So it's not possible for us to hardcode that.

websitetoolbox.com.apexforum.json
webtoolbox: Ok. Our load balancers currently have ever-changing IP addresses, but I'm looking into having at least one static IP address that can be used here instead.

websitetoolbox.com.forum.json
webtoolbox: We've seen DNS providers reluctant to go async. We'd also like to avoid the complexity and have as many DNS providers integrated as possible.

We're working on changing the REDIR to an A record to have maximum compatibility.

webtoolbox:
Thanks for all the feedback so that this could be set up correctly. All of the changes have been made. Requesting merge...

websitetoolbox.com.forum.json
```diff
@@ -0,0 +1,30 @@
+{
+  "providerId":"websitetoolbox.com",
```
arnoldblinn, line 2: I would consider if removing syncPubKeyDomain from this template is wise.  Typically a static template (like the other one) is somewhat safe to apply without a signature.

But you have a variable in a host name here.  If I want to mess with you, I generate a link that applies this template with a random v1 variable;  and this would potentially overwrite and hijack a random sub-domain in the user's domain. Of course, it would point to your forum application.  So it is probably OK to leave out.

Just thinking out loud.

Depending on what @pawel-kow says about the variable in the host, we could publish this template.  I supsect the DNS Providers would onboard despite and allow the variable in the host....

webtoolbox: Thanks, that's a good point. We omitted it initially just to keep things as simple as possible while setting it up with one DNS provider. Once things are working properly, perhaps we could make a modification and add in syncPubKeyDomain before rolling out to other DNS providers or going live.

websitetoolbox.com.forum.json
pawel-kow: Interesting case indeed.
Generally I would argue that the template represents more than one service and this is the reason why we face this issue.
One service is the mailing service applied at @ and the other one is the website service at some host.
One can also argue the opposite, that both services are in fact the same at SPF and DKIM are there to assure e-mail sent by the website. And thinking laud I think the latter is in fact true as one wants the SPF and DKIM disappear once the website is gone.

So assuming this is a one service, we might need to reverse the logic of the template in general, following where is the mail service in this case and this is the website. Especially I can imagine the customer being able to setup more than one such website on the same domain, what with current template would lead to mutual exclusion - basically removing previous host when a new one is applied. A mix of multiInstance and essential settings would have to be used to mitigate.

One solution might be to make use of the host parameter and place the website with `@` host as a template root.
Then the SPFM and DKIM would need to take `%domain%.` and `websitetoolbox._domainkey.%domain%.` respectively as host. We somewhat specified it in the spec in 6.9.2, this also allows for conflict detection, just that nobody used it this way so far so there might be some testing/adjustments involved. For me it looks however as a right expression of the intent behind the template.

And there is one small issue here, the DKIM record would in this case always conflict with the other instance (even if in fact the records are identical I guess the implementations would rise a conflict anyway). This can be mitigated by marking this record as non-essential, then the 2nd instance will drop the 1st record instance silently.
This will work when adding services, there might be a small drawback when removing services, as one instance which happens to be assigned to DKIM record would drop it even if other instances would still need it.
To be fully covered we would need sth like shared static records, again a new feature.

Other option would be going for 2 templates, which would then either have UX disadvantage (2 times sync flow), or use of async as Arnold mentioned, or would need a spec change to have "bulk" apply call for more than one template of the same provider at once. We at IONOS would be open to this feature as we saw same need in few of internal use-cases.

The final and maybe the least relevant consideration is that whether blog.example.com shall send emails under mailer@example.com or mailer@blog.example.com. This is however the product freedom and the end-user decision in the end. Sticking to mailer@blog.example.com would however solve all of the issues above without any need to craft new spec features or functional drawbacks.

webtoolbox: Thanks for your reply. 

> One solution might be to make use of the host parameter and place the website with @ host as a template root.
Then the SPFM and DKIM would need to take %domain%. and websitetoolbox._domainkey.%domain%. respectively as host. We somewhat specified it in the spec in 6.9.2, this also allows for conflict detection, just that nobody used it this way so far so there might be some testing/adjustments involved. For me it looks however as a right expression of the intent behind the template.

This goes over my head a bit. Perhaps you can make whatever changes are needed to the template?

We would prefer that mail be sent from the client's root domain name. It's a bit atypical to be sending email from forum.mydomain.com instead of mydomain.com. However, if it's really the best/easiest way to set it up, we can still do it.

It's extremely rare for a client to set up two forums on the same domain name. In fact, I think I've never seen it with any of our clients. If they need to discuss different things, they create different categories on the same forum rather than creating an entirely new forum and paying two separate subscription fees. The only time a client has multiple forums is if they have multiple websites, in which case they use multiple different domain names. So perhaps we can avoid all of this and ignore the case where a client wants multiple forums on multiple domain names? We could get ahead of it in our own code and show a message to the client to set up the records manually or contact our support team if we see that a forum is already set up with a particular domain name and they want to set up another forum on the same domain. Then we could set up both forums on that domain manually instead of through domain connect.

We would like to move forward as fast as possible and I assume that crafting new parts of the spec would take a lot of time.

webtoolbox: Thoughts on this? Can we get this merged?

pawel-kow: Sorry, I thought I already commented here.

If for this service the case is indeed to have only one template of this kind per zone, then I am OK with the template proposed.
At least at IONOS it would to most extent work. We have currently a small bug on re-apply if you would change the variable host (the old host would be left behind).

@arnoldblinn I propose to have an announcement about this template on Slack and let a comment period of 2-3 weeks (it's vacation time) before we merge. What do you think?

arnoldblinn: Here is my take.

I don't think anyone will really care.  The DNS Providers aren't hyper thoughtful at this level of detail with variables in the host name.

I also think you'll have problems with records lingering in DNS.  1&1 and GoDaddy will remove a template before re-apply; other providers that don't maintain stateful template changes will leave the old record around.

I don't think contorting the protocol's interpretation that %domain% alone means writing a record at just the domain level (e.g. if host and domain are set, writing a record at the domain component) would fix this.  Note: I looked into this, and the example implementation does not handle this.  Could mod it, but it is fairly disruptive.

When Domain Connect was built, it was not imagined that a template could be applied to foo.example.com (host=foo and domain=example.com) and have records set on just example.com.  There is a benefit of sending the outbound mail from blog.example.com (vs. example.com); this is where the forums live.  

If this provider wants to ship the template as is, we can approve.  But they are likely going to be fighting this quirk with each DNS Provider onboarding.

webtoolbox: Thanks for the detailed feedback. It's really rare that a client changes the subdomain that they want to use. Once they already have links to that subdomain on their site and possibly in search engines, changing it means breaking those links as per our software. So it's a very rare thing. In those rare cases, if the old subdomain CNAME record still stays there for some DNS providers, it wouldn't be a deal breaker situation. In a way it might actually be good because then in the future we could add functionality to map that old subdomain to their new subdomain so that links don't break.

We would prefer this approach rather than using a subdomain for email. That's not the usual approach and it'll definitely result in clients sending in support inquiries asking if and how they can change it.

webtoolbox: Any update on this? Can we get this merged now?

arnoldblinn:
Sorry for the delay; we can merge it.  

But I'm wondering if the better approach isn't to apply the template to host=forms&domain=example.com, with the template being set to:

CNAME = %fqdn% (or @), and the other records using a host of %domain%.

In theory, I think this is to spec but we may uncover a few bugs with various DNS Providers with this approach (which in the long run I think is a good thing).

@pawel-kow ?

pawel-kow:
@arnoldblinn , isn't your proposal the same as I suggested in https://github.com/Domain-Connect/Templates/pull/136#discussion_r465480290 ?

You didn't seem to be happy about it, changed your mind?

Anyway, to be exact about the proposal here as I understand it:
```
{
  "providerId":"websitetoolbox.com",
  "providerName":"Website Toolbox",
  "serviceId":"forum",
  "serviceName":"Website Toolbox Forum",
  "version": 1,
  "logoUrl":"https://cdn.websitetoolbox.com/images/logos/wt_logo_transparent.png",
  "description":"Enables a subdomain to work with Website Toolbox.",
  "variableDescription":"host is the subdomain that wants to be used. ie: 'forum' for forum.mydomain.com.",
  "hostRequired": true,
  "records":[
    {
      "type":"CNAME",
      "host":"@",
      "pointsTo":"custom.websitetoolbox.com",
      "ttl":3600
    },
    {
      "type": "SPFM",
      "host": "%domain%.",
      "spfRules": "a:mailers.websitetoolbox.com",
      "ttl": 3600
    },
    {
      "type": "CNAME",
      "host": "websitetoolbox._domainkey.%domain%.",
      "pointsTo": "websitetoolbox._domainkey.websitetoolbox.com",
      "ttl": 3600
    }
  ]
}
```

Generally I am for it as I also think it reflects the best the semantic meaning of the template. I believe it requires some testing and maybe use of the trailing dot on the host level to note that `%domain%.` is a fully qualified name, not a subdomain relative to the template root.

arnoldblinn:
@pawel-kow It wasn't unhappy with; it was realization that it will unearth some bugs in people's implementations.  But I do think it is inline with the spec and in some ways perhaps even clarifying. And it is because of this that I gave it is a second thought.  It will be worth the effort to debug implementations to make it work.

I know that the variable in the host would likely work out the gate; but it will have other problems (leaving behind records, not doing conflict detection) that will cause bigger problems in the long run.

The open source implementation in github can be fixed relatively easily; I'm sure 1&1 and GoDaddy will make it work quickly (which is probably around 35-40% of the domains out there).  Others that use the open source will just need to update their code.  So I don't think it would be too painful....

arnoldblinn:
Sorry for the delay.  We are getting this fixed, so the proper template will be the proposal from Pawel 5 days ago.

webtoolbox:
Ok, thanks for the update.

webtoolbox:
Hi there! Any update on this? It has been about a month and we'd really like to get this live. It will be a big improvement for our clients. Thank you!

pawel-kow:
@webtoolbox , as per @arnoldblinn comment on Aug'18th we are waiting for the template update as per my proposal in the comment from on Aug'14th https://github.com/Domain-Connect/Templates/pull/136#issuecomment-673926894

webtoolbox:
Oh. The change has been made now.

websitetoolbox.com.apexforum.json
```diff
@@ -0,0 +1,36 @@
...
+      "type":"CNAME",
+      "host":"www",
+      "pointsTo":"custom.websitetoolbox.com",
+      "ttl":3600
+    },
+    {
+      "type": "SPFM",
+      "host": "@",
+      "spfRules": "a:mailers.websitetoolbox.com",
+      "ttl": 3600
```
arnoldblinn, line 27: TTL isn't needed on SPFM

PR #142, opened 2020-09-26
Changed domain variable to spdomain

webtoolbox:
A domain provider already internally uses the %domain% variable, so we have changed it to %spdomain%.


pawel-kow:
Well, the template was actually meant to leverage on the built in `%domain%` variable. This should be supported as per spec. The new variable will have the same value anyway.

PR #145, opened 2020-10-19
Adding Bluehost.com DomainConnect integration

csjewell:


bluehost.com.hosting.json
```diff
@@ -0,0 +1,167 @@
...
+      "ttl": 14400,
+      "priority": 0
+    },
+    {
+      "data": "v=spf1 a mx include:websitewelcome.com ~all",
+      "type": "TXT",
+      "host": "@",
+      "ttl": 14400,
+      "txtConflictMatchingMode": "Prefix",
+      "txtConflictMatchingPrefix": "v=spf1"
```
arnoldblinn, line 80: We prefer to use the SPFM convention for SPF records.  See the spec.

csjewell: OK. Will look that up and make the appropriate changes.

I had originally done this 4 months ago to work with Google and we're just now getting it contributed here.

csjewell: Change made in a04edc5.

pawel-kow:
I'm fine with the template after the SPFM change.

What strikes me however is that it's a large monoblock with hosting, mail and few other services. Also not grouped by any groupId.
There is an advantage of only one apply needed for this case, the disadvantage is that with a provider taking care of template integrity, like GoDaddy or IONOS, the whole template would be dropped if any of the records would need to be removed. Say someone connects O365 on a domain would trigger the A and all other records dropping as well.

Maybe it's worth a while to see which records are not essential for the template integrity and use `essential` setting accordingly.

csjewell:
Thank you - I was going to ask that you merge it as is, and then my coworkers and I can do that analysis and make the appropriate changes at a later time. (I'll create that issue to do so within our project tracker right now.)

annreilly:
Hello Pawel,

Just checking to see if we could accommodate Curtis; request to merge the bluehost template as is since it passed the check?

pawel-kow:
@annreilly , the template was merged by Arnold on Nov-4th, wasn't it?

PR #148, opened 2020-12-01
Update Google templates to set "syncBlock" to false in favor of supporting SYNC flow soon

navneetgoel-domainconnect:


pawel-kow (changes requested):
I think this change should bump the version number.

navneetgoel-domainconnect:
> I think this change should bump the version number.
Done

PR #152, opened 2021-01-11
Add three new HubSpot templates

anev:
Added a new HubSpot template with 3 records: A, CNAME, TXT

 In order to be able to support multiple patterns and to provide more flexibility


pawel-kow:
@anev, are these 3 templates for the very same service? What is the reason for 3 templates instead of 1 template with 3 groupIds?

anev:
> @anev, are these 3 templates for the very same service? What is the reason for 3 templates instead of 1 template with 3 groupIds?

Hey Pawel,
in our pipeline, we need to set 1 or several records.
I went through the documentation again and looks like groupId works for us.

If there is 1 template with 3 records, and each record has a unique groupId. In this case, we can set one, two, or three records.
Is that right? If it is right I will merge 3 templates into a single one.

PR #154, opened 2021-01-22
Adding the essential key and Bluehost sister brands.

csjewell:
The 'essential' key changes were requested back in PR#145 - finally was given the time to do them.

### DIFF
--- a/bluehost.com.hosting.json
+++ b/bluehost.com.hosting.json
@@ -14,72 +14,84 @@
       "pointsTo": "%ip%",
       "type": "A",
       "host": "@",
+      "essential": "Always",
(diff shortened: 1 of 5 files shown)


pawel-kow:
@csjewell thanks for the drop. Did you consider using `sharedProviderName` flag to cover the brands to avoid copy-cat templates? `syncRedirectDomain` can be a list so you can cover all the brand redirects as needed.

csjewell:
We thought about it, yes, and good callout to ask about it.

There are changes in there (specifically the logoUrl) that sharedProviderName didn't appear to cover.

pawel-kow:
LGTM!

PR #158, opened 2021-01-28
A record is modified to intended CNAME

davestys:
A record converted to intended CNAME record.

### DIFF
--- a/aweber.com.landing-pages-subdomain.json
+++ b/aweber.com.landing-pages-subdomain.json
@@ -6,11 +6,12 @@
    "version": 1,
    "logoUrl":"https://assets.aweber-static.com/page-templates/assets/img/powered_by.png",
    "description":"Configure a domain to AWeber Landing Pages",
-   "variableDescription":"%subdomain such as www or go%",
+   "variableDescription":"",
+   "hostRequired": true,
    "records":[
        {
          "type": "CNAME",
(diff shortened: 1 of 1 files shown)


aweber.com.landing-pages-subdomain.json
```diff
@@ -9,7 +9,7 @@
    "variableDescription":"",
    "records":[
        {
-         "type": "A",
+         "type": "CNAME",
```
pawel-kow, line 12: CNAME on APEX is not allowed. If intended usage is only for subdomains, please set `hostRequired` to `true`.

davestys: ah that's it. thanks

davestys: So, best if I use a variable like %subdomain%:
`"host": "%subdomain%"`

pawel-kow:
@davestys, actually I mean you should add `hostRequired` parameter and leave the host as it was with `@`

davestys:
ok will do.

davestys:
@pawel-kow is there anything else I need to do to have this update merged? Thanks

pawel-kow (approved):

pawel-kow:
Done

davestys:
thanks!

PR #159, opened 2021-02-02
Editing the DKIM key naming convention

davestys:
Changes based on AWeber's final decision on naming convention for the Optin.com mail sending domains.


pawel-kow:
LGTM!

PR #160, opened 2021-02-03
Removing variable  %account-subdomain% from CNAMEs

davestys:
Final edit prior to implementation


davestys:
@pawel-kow, I was able to confirm final details of the template with my team. I do not anticipate any further edits.

PR #168, opened 2021-05-19
Template for Twilio Sendgrid

cmalec:
Pre answering the common feedback section:

> 1. Misnamed templates

I don't think its misnamed

> 2. Variables too broad in scope (i.e. possible to be misused to set any arbitrary record and conflict with other template). Too broad scope example: @ TXT "%verification%". Better usage: @ TXT "foo-verification=%verification%".

We need these unique per user so variables are required

> 3. Variables as a host name. These are highly discouraged unless explicitly needed

Again needs to be unique per user

> 4. Not using digital signatures (or warnPhishing) on “insecure templates”

We will be signing

> 5. not specifying syncRedirectDomain when intended to use redirect_uri parameter in the synchronous flow

We do

> 6. not specifying syncBlock for a template only intended for asynchronous flow

N/A

> 7. not specifying hostRequired on the templates not applicable to domain APEX (with CNAME on @ host)

N/A

> 8. specifying a TXT record with SPF content (i.e. "v=spf1 ...") instead of using SPFM record type

N/A

### DIFF
--- a/sendgrid.com.domain-auth.json
+++ b/sendgrid.com.domain-auth.json
@@ -10,19 +10,19 @@
   "records":[
     {
       "type": "CNAME",
-      "host": "%host%",
+      "host": "@",
       "pointsTo": "%POINTS_TO%",
       "ttl": 1800
     },
     {
       "type": "CNAME",
-      "host": "%SUBDOMAIN_DKIM1%",
+      "host": "%SUBDOMAIN_DKIM1%.%domain%.",
       "pointsTo": "%POINTS_TO_DKIM1%",
       "ttl": 1800
     },
     {
       "type": "CNAME",
-      "host": "%SUBDOMAIN_DKIM2%",
+      "host": "%SUBDOMAIN_DKIM2%.%domain%.",
       "pointsTo": "%POINTS_TO_DKIM2%",
       "ttl": 1800
     }
--- a/sendgrid.com.domain-auth.json
+++ b/sendgrid.com.domain-auth.json
@@ -7,6 +7,7 @@
   "logoUrl":"https://sendgrid.com/brand/sg-twilio/SG_Twilio_Lockup_RGBx1.png",
   "syncPubKeyDomain": "domainconnect.sendgrid.net",
   "syncRedirectDomain": "sendgrid.com",
+  "hostRequired": true,
   "records":[
     {
       "type": "CNAME",
--- a/sendgrid.com.domain-auth.json
+++ b/sendgrid.com.domain-auth.json
@@ -1,0 +1,30 @@
+{
+  "providerId":"sendgrid.com",
+  "providerName":"Twilio SendGrid",
+  "serviceId":"domain-auth",
+  "serviceName":"Twilio SendGrid Domain Authentication",
+  "version": 1,
+  "logoUrl":"https://sendgrid.com/brand/sg-twilio/SG_Twilio_Lockup_RGBx1.png",
+  "syncPubKeyDomain": "domainconnect.sendgrid.net",
+  "syncRedirectDomain": "sendgrid.com",
+  "records":[
+    {
+      "type": "CNAME",
+      "host": "%SUBDOMAIN%",
+      "pointsTo": "%POINTS_TO%",
+      "ttl": 1800
+    },
+    {
+      "type": "CNAME",
+      "host": "%SUBDOMAIN_DKIM1%",
+      "pointsTo": "%POINTS_TO_DKIM1%",
+      "ttl": 1800
+    },
+    {
+      "type": "CNAME",
+      "host": "%SUBDOMAIN_DKIM2%",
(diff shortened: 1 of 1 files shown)


pawel-kow:
@cmalec thanks for the submission and for going thorugh the recommendations, this makes the review much easier!

The only question I'd have is about %SUBDOMAIN%, %SUBDOMAIN_DKIM1% and %SUBDOMAIN_DKIM2% variables.
Are those intended to put the template in more than one instance on different subdomains?
In this case the recommendation would be to rather use host parameter instead - this would allow for more instances of the template on different subdomains. Alternatively multiInstance setting might be needed to allow for more than one instance.

cmalec:
@pawel-kow so in my query string I would have `domain=example.com&SUBDOMAIN=em123&SUBDOMAIN_DKIM1=s1._domainkey&SUBDOMAIN_DKIM2=s2._domainkey` so I do not believe this is a multi instance case. But also I do not think i could use the host param here. Please let me know if I am misunderstanding anything or doing something incorrectly

pawel-kow:
@cmalec the question would be if you would also want to apply the same template to the same domain with SUBDOMAIN=em456.
If yes, you either need to use host or multi instance.

cmalec:
ok @pawel-kow got it I changed it to `host`. and now my query params would not contain `SUBDOMAIN` and instead that would be `host`

sendgrid.com.domain-auth.json
```diff
@@ -0,0 +1,30 @@
...
+  "serviceId":"domain-auth",
+  "serviceName":"Twilio SendGrid Domain Authentication",
+  "version": 1,
+  "logoUrl":"https://sendgrid.com/brand/sg-twilio/SG_Twilio_Lockup_RGBx1.png",
+  "syncPubKeyDomain": "domainconnect.sendgrid.net",
+  "syncRedirectDomain": "sendgrid.com",
+  "records":[
+    {
+      "type": "CNAME",
+      "host": "%host%",
```
pawel-kow, line 13: When usage of `host` is intended for subdomains then this value should be set to `@`.
Please refer to [spec](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#67-hostname-in-template) for details (also examples in [6.9.4](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#694-variable-examples).

sendgrid.com.domain-auth.json
```diff
@@ -0,0 +1,30 @@
...
+      "pointsTo": "%POINTS_TO%",
+      "ttl": 1800
+    },
+    {
+      "type": "CNAME",
+      "host": "%SUBDOMAIN_DKIM1%",
+      "pointsTo": "%POINTS_TO_DKIM1%",
```
pawel-kow, line 20: When using `host` parameter all subdomains will be re-rooted at the applied subdomain.
So if the intention is for `host=sub123` and `domain=example.com` to get a record like `s1._domainkey.example.com` please put `%SUBDOMAIN_DKIM1%.%domain%.` as a record value (note the dot at the end for FQDN).
If the intention is to generate `s1._domainkey.sub123.example.com` then it's OK like it is. AFAIK DKIM requires the record on the subdomain, but it may be specific to your implementation.

sendgrid.com.domain-auth.json
```diff
@@ -0,0 +1,30 @@
...
+      "ttl": 1800
+    },
+    {
+      "type": "CNAME",
+      "host": "%SUBDOMAIN_DKIM2%",
```
pawel-kow, line 25: Same as `%SUBDOMAIN_DKIM1%`

pawel-kow (changes requested):

sendgrid.com.domain-auth.json
cmalec: ok. all comments addressed

sendgrid.com.domain-auth.json
```diff
@@ -0,0 +1,30 @@
+{
+  "providerId":"sendgrid.com",
+  "providerName":"Twilio SendGrid",
+  "serviceId":"domain-auth",
+  "serviceName":"Twilio SendGrid Domain Authentication",
+  "version": 1,
+  "logoUrl":"https://sendgrid.com/brand/sg-twilio/SG_Twilio_Lockup_RGBx1.png",
+  "syncPubKeyDomain": "domainconnect.sendgrid.net",
+  "syncRedirectDomain": "sendgrid.com",
+  "records":[
```
pawel-kow, line 10: Please add ` "hostRequired": true,`, as now the template defines CNAME at `@` so it must be always applied with host parameter and cannot be at APEX.

cmalec: ok fixed its 🟢  now

cmalec:
thanks for the help @pawel-kow

PR #172, opened 2021-07-02
Removed unused AWeber Templates. Added a new single template for AWeb…

davestys:
Removed unused AWeber Templates. Added a new single template for AWeber with multiple groups.
Needs review from IONOS, Google Domains, GoDaddy prior to installation please.


pawel-kow:
@davestys this template as proposed in this PR is impossible, because it asks for CNAME on @ which would exclude usage on APEX domain and require setting `hostRequired` template property. You'd need at least to split into 2 templates - for APEX and for a subdomain. This is why the schema check is failing.
Also please consider if the groups of records correspond to a separate service which can be independently provisioned or deprovisioned without affecting others, or whether they constitute one service.
In the first case a separate template is recommended, in the latter one template with groupIds would be more suited.

aweber.com.email-web.json
```diff
@@ -0,0 +1,91 @@
...
+       {
+         "groupId": "web",
+         "type": "CNAME",
+         "host": "www",
+         "pointsTo": "hosted-content.aweber.com",
+         "ttl": 3600
+       },
+       {
+         "groupId": "web-subdomain",
+         "type": "CNAME",
```
pawel-kow, line 85: this record is failing the schema without `hostRequired` template property set

pawel-kow (changes requested):

aweber.com.email-web.json
davestys: I created a separate template for the subdomain situation. Decision is made to maintain this as one service where one or more groups can be applied at once.

davestys:
Ok, looks like the template is valid @pawel-kow . Thanks for the advice. Let me know if you see any further concerns.

pawel-kow (approved):
All good here.

pawel-kow:
@rphilip can you revuew for GoDaddy? Not sure who from Google Domains. @davestys do you have a contact person there to ask? I will post on Slack on this anyway.

davestys:
> @davestys do you have a contact person there to ask? I will post on Slack on this anyway.

Yes, Xiaotong Chen. I will send an email asking for review as well.

wdbaker-godaddy (approved):
No problems found.

PR #173, opened 2021-07-08
Add verification TXT to GoDaddy and SecureServer Professional Email Templates

fgorodishter-godaddy:


wdbaker-godaddy (approved):
Approved

PR #177, opened 2021-08-02
Add Public Internet hosting templates

tombeard:
I hope these are acceptable.  I've been through the common feedback points and I think I've addressed them all.

- Misnamed templates
I've followed the naming scheme

- Variables too broad in scope (i.e. possible to be misused to set any arbitrary record and conflict with other template). Too broad scope example: @ TXT "%verification%". Better usage: @ TXT "foo-verification=%verification%".
No broad scope variables are used.

- Variables as a host name. These are highly discouraged unless explicitly needed
%domain% is used a host variable pblin.net.website-cname.json as this template will be used with a hostname and the SPF record wants to be on the domain root.  It's my understanding that @ will resolve to %fqdn% rather than %domain% in this case.  Please let me know if I've misunderstood.

- Not using digital signatures (or warnPhishing) on “insecure templates”
Digital signatures are used on all templates

- Not specifying syncRedirectDomain when intended to use redirect_uri parameter in the synchronous flow
syncRedirectDomain is specified on all templates

- Not specifying syncBlock for a template only intended for asynchronous flow
No asynchronous flow implementation at this time

- Not specifying hostRequired on the templates not applicable to domain APEX (with CNAME on @ host)
hostRequired is specified on the only template where a CNAME could occur on the domain root.

- Specifying a TXT record with SPF content (i.e. "v=spf1 ...") instead of using SPFM record type
SPFM is used in place of TXT for SPF records


pawel-kow:
Looks good, but no clue why Travis CI is not starting on this PR.
I'll revisit tomorrow if it fixes by itself.

pawel-kow (approved):
LGTM

pawel-kow:
Close+reopen to trigger Travis CI hopefully 🤞

tombeard:
Thanks @pawel-kow, that seemed to trigger Travis.  I've fixed the incorrect serviceId that was causing the failed check and the build is now passing.

PR #178, opened 2021-08-24
Twilio Sendgrid Domain Auth Manual Security Template

cmalec:
I am back with another template.
Pre answering the common feedback section:

> 1. Misnamed templates

I don't think its misnamed

> 2. Variables too broad in scope (i.e. possible to be misused to set any arbitrary record and conflict with other template). Too broad scope example: @ TXT "%verification%". Better usage: @ TXT "foo-verification=%verification%".

We need these unique per user so variables are required

> 3. Variables as a host name. These are highly discouraged unless explicitly needed

I learnt from last time and used host required and @ where I could.

> 4. Not using digital signatures (or warnPhishing) on “insecure templates”

We will be signing

> 5. not specifying syncRedirectDomain when intended to use redirect_uri parameter in the synchronous flow

We do

> 6. not specifying syncBlock for a template only intended for asynchronous flow

N/A

> 7. not specifying hostRequired on the templates not applicable to domain APEX (with CNAME on @ host)

hostRequired = true

> 8. specifying a TXT record with SPF content (i.e. "v=spf1 ...") instead of using SPFM record type

We have to use TXT.

### DIFF
--- a/sendgrid.com.domain-auth-manual.json
+++ b/sendgrid.com.domain-auth-manual.json
@@ -1,0 +1,31 @@
+{
+  "providerId":"sendgrid.com",
+  "providerName":"Twilio SendGrid",
+  "serviceId":"domain-auth-manual",
+  "serviceName":"Twilio SendGrid Domain Authentication (Manual)",
+  "version": 1,
+  "logoUrl":"https://sendgrid.com/brand/sg-twilio/SG_Twilio_Lockup_RGBx1.png",
+  "syncPubKeyDomain": "domainconnect.sendgrid.net",
+  "syncRedirectDomain": "sendgrid.com",
+  "hostRequired": true,
+  "records":[
+    {
+      "type": "MX",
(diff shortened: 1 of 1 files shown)


pawel-kow:
@cmalec 

>> specifying a TXT record with SPF content (i.e. "v=spf1 ...") instead of using SPFM record type
> We have to use TXT.
Why would you not be happy with SPFM? It's not critical for a subdomain template where other SPF records are less likely.... but out of curiosity?

cmalec:
@pawel-kow I think the SPF record type has been deprecated the the RFC states to use TXT records. I can't find any mention of a SPFM record anywhere but on the DomainConnect page. Also whats really the show stopper either way is we have systems in place that are looking for this TXT record and expect it to be TXT so to keep things working smoothly and in a timely manner I need that to continue. It would require multiple teams and a large amount of regression testing, that would not be possible soon.

PR #180, opened 2021-09-02
Onboarding iCloud Custom Email Domain template

rohitkandhari:


rphilip (changes requested):
CNAME priority for the last record on this template is not required.  It is also not in the integrated template with GoDaddy.  Everything else looks good. @rohitkandhari

rphilip:
+1 from me.  Can @pawel-kow  have a look and merge if all is well.  GoDaddy has integrated this template already.

pawel-kow:
@rohitkandhari can you comment on the SPF record please, why it is a TXT record instead of SPFM?
This way it may be breaking other services attached to the same domain. see https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#6102-multiple-services for details.

I also see `"multiInstance": true,` on this template. Can you elaborate how it is supposed to be used as it is not possible to multi-instance the whole template because MX records are exclusive.

rohitkandhari:
hey @pawel-kow my understanding of multiInstance was to apply the same template again in case of failures? is that incorrect?

pawel-kow:
> hey @pawel-kow my understanding of multiInstance was to apply the same template again in case of failures? is that incorrect?
Not really. For this it's enough to have a regular one-instance template.
More on `multiInstance` in the spec https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#64-apply-re-apply-and-multi-instance

icloud.com.custom_email_domain.json
```diff
@@ -0,0 +1,50 @@
+{
+   "providerId":"icloud.com",
+   "providerName":"iCloud",
+   "serviceId":"custom_email_domain",
+   "serviceName":"Custom Email Domain",
+   "version": 1,
+   "logoUrl":"",
+   "description":"This is a template to enable iCloud Custom Email Domain",
+   "syncPubKeyDomain": "icloudmailadmin.com",
+   "multiInstance": true,
```
pawel-kow, line 10: Likely not needed

icloud.com.custom_email_domain.json
```diff
@@ -0,0 +1,50 @@
...
+      {
+         "groupId":"icloud-verification",
+         "type":"TXT",
+         "host":"@",
+         "data":"%icloud_token%",
+         "ttl":3600
+      },
+      {
+         "groupId":"icloud-spf",
+         "type":"TXT",
```
pawel-kow, line 21: SPF records shall normally use SPFM record type.
Usage of TXT record shall be an exception and avoided as far as possible.

pawel-kow (changes requested):

pawel-kow:
@rohitkandhari  any progress on this PR?

icloud.com.custom_email_domain.json
pawel-kow: @rohitkandhari thanks for correcting other issues.
How about this last one with the use of SPFM?

I understand the convenience of using `redirect` SPF directive, but still it is not compatible with other usages of a domain, which can be well configured in parallel to iCloud. Especially if the domain hosts a website it is essential that the CMS is able to send out emails so the rules have to be merged and not exclusive to iCloud.
Details in the [specification](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#multiple-services).

A pure TXT record as proposed here may also put a second SPF record if one exists already, rendering invalid SPF configuration.

Also any verification schema which checks exactly the content of SPF must not be employed, because SPF can be changed any time to accomplish desired customer configuration.

pawel-kow: @rohitkandhari I appreciate a lot your update to SPFM. 

One operational remark you may want to consider would be to use different variable, like `%spfm_data%` in the v2 template, so that you can serve both providers still using v1 template and the ones using v2 template same time, by sending `spf_data` and `spfm_data` variables in parallel for the transition period until all providers switch to v2 template.

I can also approve and merge this template as it is if you confirm you would not consider my suggestion.

rohitkandhari: So what I have done today is check the template version which is returned back from the supported call https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#query-supported-template and then publish the data accordingly.  Is this good or there can be providers who do not send the version back and we can send both the spf_data and spfm_data and the provider would substitute the appropriate value depending upon the template supported?

pawel-kow: Not all providers reports back template version AFAIK but if your test shows you get the version back from all providers who deployed the template then indeed you are fine as well. Please let me know  if to merge as is.

pawel-kow (approved):

icloud.com.custom_email_domain.json
rohitkandhari: I am going to make few changes to incorporate what we discussed above. Will update one ready so that we can merge the PR.

rohitkandhari: hey @pawel-kow we are good to merge this, thanks for all the good feedback

PR #181, opened 2021-09-24
Short.io templates

gugu:
These templates are for connecting clients' domains to Short.io link shortener. We have 2 IPs in case of routing or network issues and latency-aware CNAME record if it is possible


pawel-kow:
LGTM

PR #184, opened 2021-10-01

GokulnathMadhivanan:
We are adding DMARC record entry in template.

### DIFF
--- a/zoho.com.zmail_hosting.json
+++ b/zoho.com.zmail_hosting.json
@@ -53,6 +53,13 @@
         "host": "%dkim_host%._domainkey",
         "data": "%dkim_data%",
         "ttl": 3600
+    },
+    {
+        "groupId":"zoho-dmarc",
+        "type": "TXT",
+        "host": "_dmarc",
+        "data": "%dmarc_data%",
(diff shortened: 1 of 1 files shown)


zoho.com.zmail_hosting.json
```diff
@@ -53,6 +53,13 @@
         "host": "%dkim_host%._domainkey",
         "data": "%dkim_data%",
         "ttl": 3600
+    },
+    {
+        "groupId":"zoho-dmarc",
+        "type": "TXT",
```
pawel-kow, line 59: DMARC is a bit tricky, as it actually affects the behavior of the whole domain.
I think the template should assure, that there is no DMARC record present already and also allow for removal of DMARC without dropping the whole template.

So I propose to add `txtConflictMatchingMode=All` to assure no duplicates and `essential=OnApply` to allow removal or change of the record.

pawel-kow (changes requested):

pawel-kow:
@GokulnathMadhivanan you should also bump the version to "2".

pawel-kow (approved):
Looks OK now.

PR #186, opened 2021-10-11
Gearlaunch Template

joaogoulart-xogito:
Hello everyone,

This is the template for the domain connect, first we are trying to connect with GoDaddy. 

We will need two variables that our infrastructure will provide programmatically and then we will pass it as variables in order to add these CNAME's values on DNS from the client.

I'm not 100% sure about the syncPubKeyDomain field, should I use it? Will it affect to test it on dev and stg environments?

### DIFF
--- a/gearlaunch.com.website.json
+++ b/gearlaunch.com.website.json
@@ -1,0 +1,24 @@
+{  
+   "providerId":"gearlaunch.com",
+   "providerName":"gearlaunch",
+   "serviceId":"website",
+   "serviceName":"Gearlaunch Site",
+   "version": 1,
+   "syncPubKeyDomain" : "Gearlaunch.com",
+   "description":"Enables a domain to work with Gearlaunch",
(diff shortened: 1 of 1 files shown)


joaogoulart-xogito:
@pawel-kow thank you for your response. 

How can I specify multiple domains under syncPubKeyDomain? On spec I did not find any example using multiple domains, can I just do a comma separe list?

pawel-kow:
@joaogoulart-xogito generally the risk exposure with a phishing-friendly template is on the DNS provider side, not on your side as a service provider.
The second CNAME with 2 variables is dangerous in a sense, that it allows any record to be set on the host, especially it allows for many schemas of ownership verification or even issue of SSL certificate over ACME protocol or other forms of DNS verification.

At least the template should define in such case would be `warnPhishing` property, then it can be merged into the repository. This is the weakest defence [defined by the specification ](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#warnings). It will be then anyway up to each provider to decide if they onboard such template or not.

PR #191, opened 2021-11-18
Template for the.com.

stabarez:
Template for the.com users to easily set up DNS records needed to integrate.

### DIFF
--- a/goentri.com.the-com.json
+++ b/goentri.com.the-com.json
@@ -1,0 +1,32 @@
+{
+  "providerId": "goentri.com",
+  "providerName": "Entri",
+  "serviceId": "the-com",
+  "serviceName": "The.com",
+  "version": 1,
+  "logoUrl": "https://cdn.goentri.com/logo.svg",
+  "description": "Allows user to easily set up domain using Entri",
+  "syncPubKeyDomain": "goentri.com",
+  "syncRedirectDomain": "api.goentri.com, goentri.com",
+  "variableDescription": "v1 is the customer hosted domain name, sslNameVerification and sslValueVerification are to validate the certificate.",
+  "hostRequired": true,
+  "sharedProviderName": true,
+  "records": [
+    {
+      "type": "CNAME",
+      "host": "@",
+      "pointsTo": "%v1%",
+      "ttl": 3600
+    },
+    {
+      "type": "CNAME",
+      "host": "%sslNameVerification%",
+      "pointsTo": "%sslValueVerification%",
+      "ttl": 3600
+    },
+    {
+      "type": "REDIR301",
+      "target": "%target%"
+    }
+  ]
+}
--- a/goentri.com.the-com.json
+++ b/goentri.com.the-com.json
@@ -8,25 +8,25 @@
   "description": "Allows user to easily set up domain using Entri",
   "syncPubKeyDomain": "goentri.com",
   "syncRedirectDomain": "api.goentri.com, goentri.com",
-  "variableDescription": "v1 is the customer hosted domain name, sslNameVerification and sslValueVerification are to validate the certificate.",
-  "hostRequired": true,
+  "variableDescription": "webSiteHost is the customer website host, webSitePointsTo is the customer hosted domain name, sslNameVerification and sslValueVerification are to validate the certificate.",
   "sharedProviderName": true,
   "records": [
     {
       "type": "CNAME",
-      "host": "@",
-      "pointsTo": "%v1%",
+      "host": "%webSiteHost%",
+      "pointsTo": "%webSitePointsTo%",
       "ttl": 3600
     },
     {
       "type": "CNAME",
-      "host": "%sslNameVerification%",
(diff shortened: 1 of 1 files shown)


goentri.com.the-com.json
```diff
@@ -0,0 +1,32 @@
...
+      "ttl": 3600
+    },
+    {
+      "type": "CNAME",
+      "host": "%sslNameVerification%",
+      "pointsTo": "%sslValueVerification%",
+      "ttl": 3600
+    },
+    {
+      "type": "REDIR301",
```
pawel-kow, line 28: I am a bit confused by this template.
What kind of configuration would you expect?
Say setup for "example.com".
Do you expect `example.com` redirect to `www.example.com` and then CNAME on `www.example.com` + SSL validation CNAME on a host `%sslNameVerification%.example.com` pointing to `%sslValueVerification%`

stabarez: @pawel-kow From this template I'd expect (considering host=www):

Redirect example.com to www.example.com

CNAME | www | v1
CNAME | sslNameVerification.**www** | sslValueVerification

From the back and forth with Google in regards to this template we understood that when a `host` is passed it will be included on every record host. Notice also that we set `hostRequired` true.

goentri.com.the-com.json
```diff
@@ -0,0 +1,32 @@
...
+  "description": "Allows user to easily set up domain using Entri",
+  "syncPubKeyDomain": "goentri.com",
+  "syncRedirectDomain": "api.goentri.com, goentri.com",
+  "variableDescription": "v1 is the customer hosted domain name, sslNameVerification and sslValueVerification are to validate the certificate.",
+  "hostRequired": true,
+  "sharedProviderName": true,
+  "records": [
+    {
+      "type": "CNAME",
+      "host": "@",
```
pawel-kow, line 17: As per [this comment](https://github.com/Domain-Connect/Templates/pull/191#discussion_r753241156)
`www` instead of `@`

goentri.com.the-com.json
```diff
@@ -0,0 +1,32 @@
...
+      "pointsTo": "%v1%",
+      "ttl": 3600
+    },
+    {
+      "type": "CNAME",
+      "host": "%sslNameVerification%",
```
pawel-kow, line 23: As per [this comment](https://github.com/Domain-Connect/Templates/pull/191#discussion_r753241156)
`%sslNameVerification%.www` instead of `%sslNameVerification%`

goentri.com.the-com.json
pawel-kow: here add host `@`

goentri.com.the-com.json
```diff
@@ -0,0 +1,32 @@
...
+  "providerName": "Entri",
+  "serviceId": "the-com",
+  "serviceName": "The.com",
+  "version": 1,
+  "logoUrl": "https://cdn.goentri.com/logo.svg",
+  "description": "Allows user to easily set up domain using Entri",
+  "syncPubKeyDomain": "goentri.com",
+  "syncRedirectDomain": "api.goentri.com, goentri.com",
+  "variableDescription": "v1 is the customer hosted domain name, sslNameVerification and sslValueVerification are to validate the certificate.",
+  "hostRequired": true,
```
pawel-kow, line 12: This won't be needed.

pawel-kow (changes requested):

goentri.com.the-com.json
pawel-kow: OK, so kind of as I expected.
The issue is that with this approach the redirect wouldn't be applied to the main domain, because `host` would be appended also here.

I added review comments to indicate needed changes.

stabarez: @pawel-kow Got it, from the docs I understood that REDIR301 record type expected one single field **target**, since this won't be a record but a URL redirect.

Our intention was to include the host & domain on the **target** for consistency.

From your proposed changes below I understand **host** is hardcoded to **www** we'd like the ability to make it variable, this way users could setup any subdomain like: **www**, **blog**, etc. Is there a way to fix the redirect issue an keeping host variable?

pawel-kow: > from the docs I understood that REDIR301 record type expected one single field target, since this won't be a record but a URL redirect.
The specification of REDIR301 is quite shallow, I agree. Mainly because it's an extension and not offered by every DNS provider.

Anyway, for REDIR301 parameters like `host`, `groupId` or `essential` apply as well. I will open an issue on the spec project to cover it better.

> From your proposed changes below I understand host is hardcoded to www we'd like the ability to make it variable, this way users could setup any subdomain like: www, blog, etc. Is there a way to fix the redirect issue while keeping host variable?

you can use a variable instead of `www` if it's needed

stabarez: @pawel-kow Addressed your comments above. Lmk how it looks.

pawel-kow (approved):
LGTM!

PR #196, opened 2021-11-25
Template for Crisp Watch feature.

stabarez:
Template for Crisp Watch users to easily set up DNS records needed to integrate.


pawel-kow:
LGTM!

PR #197, opened 2021-11-26
Template for Crisp Email feature.

stabarez:
Template for Crisp Email users to easily set up DNS records needed to integrate.

### DIFF
--- a/goentri.com.crisp-email.json
+++ b/goentri.com.crisp-email.json
@@ -26,10 +26,9 @@
       "ttl": 3600
     },
     {
-      "type": "TXT",
+      "type": "SPFM",
       "host": "@",
-      "data": "v=spf1 redirect=_spf.crisp.email",
-      "ttl": 14400
+      "spfRules": "redirect=_spf.crisp.email"
     },
     {
       "type": "CNAME",
--- a/goentri.com.crisp-email.json
+++ b/goentri.com.crisp-email.json
@@ -1,0 +1,53 @@
+{
+  "providerId": "goentri.com",
+  "providerName": "Entri",
+  "serviceId": "crisp-email",
+  "serviceName": "Crisp",
+  "version": 1,
+  "logoUrl": "https://cdn.goentri.com/logo.svg",
+  "description": "Allows user to easily set up domain using Entri",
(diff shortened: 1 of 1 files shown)


goentri.com.crisp-email.json
```diff
@@ -0,0 +1,52 @@
...
+      "type": "MX",
+      "host": "@",
+      "pointsTo": "inbound.crisp.email.",
+      "priority": 10,
+      "ttl": 3600
+    },
+    {
+      "type": "SPFM",
+      "host": "@",
+      "spfRules": "redirect=_spf.crisp.email"
```
pawel-kow, line 31: generally I'd avoid `redirect` directive in favor of `include` directive.
`redirect` does not merge well with other services and may even not be supproted by all DNS providers.

pawel-kow (changes requested):

goentri.com.crisp-email.json
stabarez: @pawel-kow this is a requirement from Crisp team specifically.

I think using `redirect` is useful for them while managing multiple domains that share same email config.

I've two questions:

- Is `redirect` directive supported by Ionos?
- Is this a blocker to merge?

I understand each provider will manually review this template before deploying on their end. If some provider doesn't support this then we have a strong argument to discuss with Crisp team.

pawel-kow: `redirect` and `include` have more less the same properties in terms of maintenance. The difference is that `redirect` terminates the processing on the source while `include` allows more rules to be added (from other services).
So using `redirect` would not be in line with the specification and the purpose of SPFM. More lecture here https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#6103-spf-record-merging

The services tend to forget that a domain might be used by multiple services and all of their SPF rules need to be merged into the final SPF. This is very the case for webhosting services where SPF can be applied by the email service but also by the webspace service which is sending emails out of a CMS.

stabarez:
@pawel-kow I'm closing this PR for now as we need to sync with our customer to see if it's possible for them to use `include` instead of `redirect`.

PR #198, opened 2021-12-01
Template for Mailshake.

stabarez:
Template for Mailshake users to easily set up DNS records needed to integrate.


pawel-kow:
Looks good.

PR #199, opened 2021-12-13
Add missing s1._domainkey to record value.

stabarez:
Add missing s1._domainkey to record value.


goentri.com.fourthwall.json
```diff
@@ -32,7 +32,7 @@
     {
```
pawel-kow, line 32: Please bump the version when changing an existing template.

PR #206, opened 2022-02-04
Fourthwall service provider update

beholders-eye:
They need a more generic RR updating template, so this is an attempt
to get there. Updated a lot of placeholders to actually receive
variables and updated the variable description for that.

### DIFF
--- a/goentri.com.fourthwall.json
+++ b/goentri.com.fourthwall.json
@@ -8,104 +8,19 @@
   "description": "Allows user to easily set up domain using Entri",
   "syncPubKeyDomain": "goentri.com",
   "syncRedirectDomain": "api.goentri.com, goentri.com",
-  "variableDescription": "ip is the website server ip, zendeskverification is the text verification value for zendesk, sendgridHost is sendgrid dynamic host.",
+  "variableDescription": "ip is the website server ip, fourthwallHost is the fourthwall host name to be added",
   "sharedProviderName": true,
   "records": [
     {
       "type": "A",
-      "host": "@",
+      "host": "%fourthwallHost%",
       "pointsTo": "%ip%",
       "ttl": 3600
     },
     {
       "type": "CNAME",
-      "host": "www",
-      "pointsTo": "@",
-      "ttl": 3600
-    },
-    {
-      "type": "CNAME",
-      "host": "%sendgridHost%.support",
-      "pointsTo": "u10250786.wl037.sendgrid.net.",
-      "ttl": 3600
-    },
-    {
-      "type": "CNAME",
-      "host": "s1._domainkey.support",
-      "pointsTo": "s1.domainkey.u10250786.wl037.sendgrid.net.",
-      "ttl": 3600
-    },
-    {
-      "type": "CNAME",
-      "host": "s2._domainkey.support",
-      "pointsTo": "s2.domainkey.u10250786.wl037.sendgrid.net.",
-      "ttl": 3600
-    },
-    {
-      "type": "CNAME",
-      "host": "zendesk1._domainkey.support",
-      "pointsTo": "zendesk1._domainkey.zendesk.com.",
-      "ttl": 3600
-    },
-    {
-      "type": "CNAME",
-      "host": "zendesk2._domainkey.support",
-      "pointsTo": "zendesk2._domainkey.zendesk.com.",
-      "ttl": 3600
-    },
-    {
-      "type": "CNAME",
-      "host": "zendesk1.support",
-      "pointsTo": "mail1.zendesk.com.",
-      "ttl": 3600
-    },
-    {
-      "type": "CNAME",
-      "host": "zendesk2.support",
-      "pointsTo": "mail2.zendesk.com.",
-      "ttl": 3600
-    },
-    {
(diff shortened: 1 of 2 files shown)


goentri.com.fourthwall.json
```diff
@@ -8,102 +8,102 @@
...
-      "host": "@",
+      "host": "%fourthwallHost%",
       "pointsTo": "%ip%",
       "ttl": 3600
     },
     {
       "type": "CNAME",
-      "host": "www",
-      "pointsTo": "@",
+      "host": "%redirectValue%",
```
pawel-kow, line 22: Not quite sure what would come into this host variable?

goentri.com.fourthwall.json
```diff
@@ -8,102 +8,102 @@
   "description": "Allows user to easily set up domain using Entri",
   "syncPubKeyDomain": "goentri.com",
   "syncRedirectDomain": "api.goentri.com, goentri.com",
-  "variableDescription": "ip is the website server ip, zendeskverification is the text verification value for zendesk, sendgridHost is sendgrid dynamic host.",
+  "variableDescription": "ip is the website server ip, zendeskverification is the text verification value for zendesk, sendgridHost is sendgrid dynamic host., fourthwallHost is the fourthwall host name to be added, supportHost is the support host being added by fourthwall",
```
pawel-kow, line 11: Looking at this template now with variables on 2 different hosts I see even more prominent that these are in fact 2 services.

A website on `%fourthwallHost%` and an email service on `%supportHost%`.
With the use of variables are they are now, instead of `host` parameter, the limitation will be that this template would be able to apply only once per zone, also when `%fourthwallHost%` and `%supportHost%` will be different.
`multiInstance` may be a way to workaround this limitation, however it would not employ this parameter as per its intended use.

An alternative would be to split into separate services for web and email. 2 templates with one consent is currently only possible with async flow, otherwise the sync flow would have to be initiated twice. Async flow was alway considered as the way to go for complex use-cases and therefore more powerful compared to the sync flow.

2 templates would offer also some flexibility, like removal of one without kicking of the other.

goentri.com.fourthwall.json
```diff
@@ -8,104 +8,19 @@
...
   "syncPubKeyDomain": "goentri.com",
   "syncRedirectDomain": "api.goentri.com, goentri.com",
-  "variableDescription": "ip is the website server ip, zendeskverification is the text verification value for zendesk, sendgridHost is sendgrid dynamic host.",
+  "variableDescription": "ip is the website server ip, fourthwallHost is the fourthwall host name to be added",
   "sharedProviderName": true,
   "records": [
     {
       "type": "A",
-      "host": "@",
+      "host": "%fourthwallHost%",
```
pawel-kow, line 16: this variable shall be now changed with `@`. The host will be then controlled by the `host` parameter when applying the template (and effectively replace `@`).

goentri.com.fourthwall.json
pawel-kow: Is this variable to take "www" or alike? `%redirectValue%` may be a bit misleading as a name. `%redirectSubdomain%` is likely more appropriate.

goentri.com.fourthwall.json
```diff
@@ -8,104 +8,19 @@
...
-      "priority": 10,
-      "ttl": 3600
-    },
-    {
-      "type": "MX",
-      "host": "support",
-      "pointsTo": "mx2.improvmx.com.",
-      "priority": 10,
+      "host": "%redirectValue%",
+      "pointsTo": "%fourthwallHost%",
```
pawel-kow, line 23: This one also changes to `@`

goentri.com.fourthwall.subdomain.json
```diff
@@ -0,0 +1,100 @@
...
+   "logoUrl": "https://cdn.goentri.com/logo.svg",
+   "description": "Allows user to easily set up domain using Entri",
+   "syncPubKeyDomain": "goentri.com",
+   "syncRedirectDomain": "api.goentri.com, goentri.com",
+   "variableDescription": "sendgridHost is sendgrid dynamic host, zendeskverification is the text verification value for zendesk, supportHost is the support host being added by fourthwall",
+   "sharedProviderName": true,
+   "records": [
+      {
+         "type": "CNAME",
+         "host": "%sendgridHost%.%supportHost%",
```
pawel-kow, line 16: Same here, `%supportHost%` can be removed in all places (when it's subdomain) or replaced with `@`

pawel-kow (changes requested):

goentri.com.fourthwall.subdomain.json
```diff
@@ -0,0 +1,100 @@
+{
+   "providerId": "goentri.com",
+   "providerName": "Entri",
+   "serviceId": "fourthwall",
```
pawel-kow, line 4: Shall be `fourthwall.subdomain` to match the file name.

goentri.com.fourthwall.json
beholders-eye: Got it, makes sense. I'm opening another PR, with the correct changes, but focused only on a "support subdomain" template. I'll skip changing this one.

PR #207, opened 2022-02-08
Add HelpDesk.com custom domain template

jfedyczak:
Hello, please add following template to the repository. Template is passing JSON schema validation.

### DIFF
--- a/helpdesk.com.custom-domain.json
+++ b/helpdesk.com.custom-domain.json
@@ -1,0 +1,39 @@
+{
+  "providerId": "helpdesk.com",
+  "providerName": "HelpDesk",
+  "serviceId": "custom-domain",
+  "serviceName": "HelpDesk custom domain setup",
+  "version": 1,
+  "logoUrl": "https://helpdesk.design/images/helpdesk/DIGITAL%20%28RGB%29/PNG/Logo_RGB_Green.png",
+  "description": "Verifies domain ownership and adds records to allow using the domain in HelpDesk",
+  "variableDescription": "helpdeskId: customer's ID in HelpDesk; dkimKeyName: DKIM key name; dkimKeyValue: DKIM key value",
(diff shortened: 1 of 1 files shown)


helpdesk.com.custom-domain.json
```diff
@@ -0,0 +1,39 @@
...
+  "version": 1,
+  "logoUrl": "https://helpdesk.design/images/helpdesk/DIGITAL%20%28RGB%29/PNG/Logo_RGB_Green.png",
+  "description": "Verifies domain ownership and adds records to allow using the domain in HelpDesk",
+  "variableDescription": "helpdeskId: customer's ID in HelpDesk; dkimKeyName: DKIM key name; dkimKeyValue: DKIM key value",
+  "syncBlock": false,
+  "sharedProviderName": false,
+  "sharedServiceName": false,
+  "syncPubKeyDomain": "helpdesk.com",
+  "syncRedirectDomain": "app.helpdesk.com,app.labs.helpdesk.com",
+  "multiInstance": true,
```
pawel-kow, line 15: why `multiInstance=true`?
the CNAME on `"host": "pm-bounces"` would always render conflicts between instances, so effectively there will be only one (unless 2nd instance is only with groupId `ownershipVerification`)

pawel-kow (changes requested):

helpdesk.com.custom-domain.json
jfedyczak: In theory a domain could be connected to two customer accounts each with separate domain key and verification record. The CNAME for pm-bounces could be common for the two installations. But if this poses a problem, I can change multiInstance to false.

jfedyczak: Changed.

pawel-kow (approved):

PR #210, opened 2022-02-23
Exly template Changes

exlyinstance:
Exly is a website hosting solution used for building websites for uses and management of users. Request for merge as custom domain linking needs to be enabled via domainconnect.


godaddy.com.exly_hosting_dev.json
pawel-kow, line 1: File name godaddy.com.exly_hosting_dev.json not matching the pattern. Expected godaddy.com.exly.json

godaddy.com.exly_hosting_prod.json
pawel-kow, line 1: same here

pawel-kow (changes requested):

PR #211, opened 2022-03-10
Add new template for WPEngine

serhiy-v:
Add new CNAME template for WPEngine


pawel-kow:
@serhiy-v what is the relation to  the https://github.com/Domain-Connect/Templates/blob/master/wpengine.com.cname.json template?
Also your account does not show any relation to WPEngine organization so I would like someone with such association to confirm this PR is legit.

daniel-savo:
@pawel-kow this is a legitimate PR

PR #213, opened 2022-03-26
Add nicesite template

diogeneshamilton:
Hey! 

We're working on a super simple website builder and we're trying to change the nameservers here because we're using Netlify to delegate to our users' respective sites.

Just curious, does Google Domains or GoDaddy support NS records through Domain Connect? Is there a list of support anywhere? Thanks!

### DIFF
--- a/nicesite.so.hosting.json
+++ b/nicesite.so.hosting.json
@@ -1,12 +1,11 @@
 {
   "providerId": "nicesite.so",
   "providerName": "nicesite",
-  "serviceId": "website",
+  "serviceId": "hosting",
   "serviceName": "nicesite",
   "version": 1,
   "logoUrl": "https://ucarecdn.com/541766e8-3c80-41b2-b3ad-5a3c3da88806/nicesite_logo_circlecrop.png",
-  "description": "Enables web hosting for nicesite domains.",
-  "variableDescription": "v1 is the apex domain",
+  "description": "Enables custom domain connection for hosting nicesite websites.",
   "records": [
     {
       "type":"A",
@@ -17,7 +16,7 @@
     {
       "type":"CNAME",
       "host":"www",
-      "pointsTo": "%v1%",
+      "pointsTo": "@",
       "ttl": 60
     }
   ]
--- a/nicesite.so.hosting.json
+++ b/nicesite.so.hosting.json
@@ -1,0 +1,32 @@
+{
+  "providerId": "nicesite.so",
+  "providerName": "nicesite",
+  "serviceId": "website",
+  "serviceName": "nicesite",
+  "version": 1,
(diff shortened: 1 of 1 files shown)


pawel-kow:
@diogeneshamilton I'm not really aware of any DNS provider supporting nameserver change. This is really an use-case for registrar, not DNS provider.
Generally speaking change of nameserver to connect to a service is really a bad pattern as it claims for exclusivity on a domain for a given service.
Domain Connect is actually a mean to avoid it, to have even better user experience and still keep DNS configuration open for other services.

diogeneshamilton:
@pawel-kow thanks, makes total sense. We explored a different path that doesn't require nameservers, so updated the file accordingly.

nicesite.so.hosting.json
```diff
@@ -0,0 +1,24 @@
...
+    {
+      "type":"A",
+      "host":"@",
+      "pointsTo":"75.2.60.5",
+      "ttl": 60
+    },
+    {
+      "type":"CNAME",
+      "host":"www",
+      "pointsTo": "%v1%",
```
pawel-kow, line 20: if `%v1%` is the APEX domain, you may also use `@` instead of variable.

nicesite.so.hosting.json
```diff
@@ -0,0 +1,24 @@
+{
+  "providerId": "nicesite.so",
+  "providerName": "nicesite",
+  "serviceId": "website",
```
pawel-kow, line 4: Automatic check failed because of file name mismatch to serviceId:
`AssertionError: File name nicesite.so.hosting.json not matching the pattern. Expected nicesite.so.website.json`
Please either rename the file or adjust serviceId accordingly

See README.md:
> Template Naming Format
> Templates should be named according the following pattern: providerId.serviceId.json

nicesite.so.hosting.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": 1,
+  "logoUrl": "https://ucarecdn.com/541766e8-3c80-41b2-b3ad-5a3c3da88806/nicesite_logo_circlecrop.png",
+  "description": "Enables web hosting for nicesite domains.",
+  "variableDescription": "v1 is the apex domain",
+  "records": [
+    {
+      "type":"A",
+      "host":"@",
+      "pointsTo":"75.2.60.5",
+      "ttl": 60
```
pawel-kow, line 15: Sure about such short TTL? 3600 is more typical for a website. It may negatively affect performance of DNS resolution.
It's just a remark, this is not blocking template merge.

pawel-kow (changes requested):

nicesite.so.hosting.json
diogeneshamilton: I updated to 300, which is same as Webflow's- we're hoping the changes are speedier than 3600.

pawel-kow:
LTGM!
Please reach out to DNS providers for the integration as per contact data on this page: https://www.domainconnect.org/dns-providers/

PR #214, opened 2022-03-30
Mailshake SPF host update

Thekkedath:


goentri.com.mailshake.json
```diff
@@ -19,7 +19,7 @@
     },
     {
       "type": "SPFM",
-      "host": "@",
+      "host": "spf",
```
pawel-kow, line 22: sure about it?
for example.com it would create a TXT record with SPF on spf.example.com.
Effectively it won't affect example.com unless this one contains an appropriate SPF with include and/or redirect directive.

Thekkedath: this is the DNS record that is being send by customer, so the above template matches this right?
```
{
        "value": "v=spf1 include:_spf.google.com ~all",
        "host": "spf",
        "ttl": 3600,
        "type": "TXT"
      }
```

pawel-kow: Yes, the template matches it just fine. It would just not work as SPF specification requires the record on APEX.
I am just OK to accept this change if they know what they are doing.
DMARC is on `_dmarc` host just fine  because this is how the specification defines it.

Thekkedath: @pawel-kow thank you for your inputs on this. we notified the customer about this and they will make the changes on their side, so closing this PR. Thanks again

PR #221, opened 2022-04-20
Add syncRedirectDomain for nicesite 

diogeneshamilton:
We're working w/ Entri so adding their domains for redirects.

### DIFF
--- a/nicesite.so.hosting.json
+++ b/nicesite.so.hosting.json
@@ -3,7 +3,7 @@
   "providerName": "nicesite",
   "serviceId": "hosting",
   "serviceName": "nicesite",
-  "version": 1,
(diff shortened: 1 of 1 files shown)


pawel-kow:
@diogeneshamilton do you want to bump the version, or it was not used in production anywhere so far?

pawel-kow:
@diogeneshamilton as there is https://github.com/Domain-Connect/Templates/blob/master/goentri.com.nicesite.json template already integrated for Entri, would it make sense to kill this PR or maybe even remove this template at all if not needed?

diogeneshamilton:
Closed this PR, I can also open another to remove the template (or feel free to go ahead and remove yourself)

PR #226, opened 2022-04-26
changes requested by godaddy

Thekkedath:


goentri.com.clickfunnels.json
```diff
@@ -10,7 +10,6 @@
     "syncRedirectDomain": "api.goentri.com, goentri.com, entri.com",
     "variableDescription": "verificationValue is the value that is used to verify domain ownership,clusterValue is the cluster assigned to the domain,websiteId is the id of website, keyValue is the generated key value",
     "sharedProviderName": true,
-    "hostRequired":true,
```
pawel-kow, line 13: I don't get this change. 
`hostRequired` is necessary due to CNAME on `@` host.

pawel-kow (changes requested):
On clickfunnels please change '%domain%' to '%domain%.' for the template to be correct as per spec an place the record on APEX.

goentri.com.clickfunnels.json
Thekkedath: only reason why I removed the `hostRequired` in the clickfunnel template is because for the TXT record instead of `@` the records being added are `entritest1.com.www.entritest1.com`. I think if we address your comment above we will be good. will make these changes  rn. Thank you

pawel-kow (approved):

PR #228, opened 2022-05-03
Bump version of Microsoft O365 template (to account for recent addition of DKIM group)

alinconstantin:
Bump version of Microsoft O365 template (to account for recent addition of DKIM group)


pawel-kow (approved):
LGTM!

PR #242, opened 2022-06-06

GOKULNATH-MADHIVANAN:
Added new fields, decreased TTL value and implemented Sync flow.


GOKULNATH-MADHIVANAN:
While request pull from zohocorp account

PR #243, opened 2022-06-08

GokulnathMadhivanan:
Added new fields, decreased TTL value and implemented Sync flow.


zoho.com.zmail_hosting.json
```diff
@@ -8,60 +8,80 @@
     "variableDescription": "",
     "syncBlock": true,
```
pawel-kow, line 9: If the intention of the template change was to unblock the sync flow, then pls change this setting to false (or remove).

pawel-kow (changes requested):

pawel-kow (approved):
LGTM!

PR #249, opened 2022-06-24
edits to split DKIM key

Thekkedath:
The DKIM key clickfunnels is using is greater than 255 characters in length which is causing some issues on google-domains and as per the suggestion from google-domains we are making some changes to split the DKIM key into multiple values with same host


Thekkedath:
Closing the PR as this was an issue on google-domain's side and not related to the length of the record as notified them.

PR #253, opened 2022-07-15
Kitkaton Storefront

miguelcabgil:


kitkaton.com.storefront.json
```diff
@@ -0,0 +1,26 @@
...
+   "records":[
+      {
+         "type":"CNAME",
+         "host":"%vfyhost%",
+         "pointsTo":"%vfyv%.acm-validations.aws.",
+         "ttl":3600
+      },
+      {
+         "type":"CNAME",
+         "host":"%hostname%",
```
pawel-kow, line 21: If the template is intended to be used on multiple subdomains the right approach is to place the records with `"host": "@"` and then use `host` parameter during the apply phase.
As far as I can see in the [AWS documentation](https://docs.aws.amazon.com/acm/latest/userguide/dns-validation.html) this will also assure the verification record placed on the right subdomain. `%vfyhost%` variable would only need to pass the `xN` record name label.

pawel-kow (changes requested):

kitkaton.com.storefront.json
miguelcabgil: Hey Pawel. The template is intended to be used in two CNAME records. The first one is to verify the domain authenticity by AWS having name and subdomain, and the second one is to store the subdomain pointing to a cloudfront url. Both records i dynamic, i mean, every time i request a custom domain creation, AWS returns different records. Below i attach the AWS documentation:
https://docs.aws.amazon.com/amplify/latest/userguide/to-add-a-custom-dom
CNAME
ain-managed-by-a-third-party-dns-provider.html

pawel-kow: Sure, got it. Still the `host` query parameter is intended to apply a template at a right subdomain.
So in your example you would use `shops` in `host` parameter and with the second CNAME specified with `"host": "@"` in the template it will render `shops.themondaysale.com` as intended.
The tricky part is about the verification record. If it is required to always place it on the APEX, then this template is correct with a limitation that only one subdomain can be ever configured for the same APEX domain.
If the verification can be also done on a subdomain level, then you are better off making changes as suggested, then you end up with an option to apply the same template several times with different subdomains without any issue.

pawel-kow: yes

pawel-kow (approved):

kitkaton.com.storefront.json
```diff
@@ -0,0 +1,26 @@
+{  
+   "providerId":"kitkaton.com",
```
pawel-kow, line 2: Please add `"hostRequired": true` a obviously CNAME is not allowed on the APEX

pawel-kow (changes requested):

PR #254, opened 2022-07-19
elementor.com email template

asafdl:


shai-shwartz:
@pawel-kow can you check, please?

pawel-kow:
Hi @shai-shwartz , I was wondering about this template with 3 verification CNAMEs. Quite unusual.
Can you tell more the use-case you are following? Would you be really putting all those 3 records at the same time? You did not use any groups, so all 3 are required.

shai-shwartz:
Hi Pawl,
We are Elementor.com a web hosting service and we allow our customers to
send emails from their custom domain by authenticating them via Sendgrid.
Sendgrid protocol returns 3 addresses and requires inserting them in their
providers so Sendgrid will be able to validate them(
https://docs.sendgrid.com/ui/account-and-settings/how-to-set-up-domain-authentication
).
We want to ease this process by using this template that enters 3 CNames.

Let me know if you have any questions
Thanks,
Shai

PR #255, opened 2022-07-24
sync pub key signature on all records

asafdl:


asafdl:
Google asked us to sign all records, even ones without params

asafdl:
@pawel-kow can you take a short look?

rphilip:
@bdickson1-godaddy @wdbaker-godaddy   Can you please take a look at this

PR #258, opened 2022-08-11
Entri edits

Thekkedath:
Changes to clickfunnels and the.com template

### DIFF
--- a/goentri.com.clickfunnels.json
+++ b/goentri.com.clickfunnels.json
@@ -13,7 +13,7 @@
   "records": [
     {
       "type": "TXT",
-      "host": "%domainapex%.",
+      "host": "@",
       "data": "clickfunnels-domain-verification=%verificationValue%",
       "ttl": 3600
     },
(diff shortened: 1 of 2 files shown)


wdbaker-godaddy (commented):
Should version be incremented to 2 for clickfunnels update?

Thekkedath:
@wdbaker-godaddy thank you for highlighting that. It was actually a miscommunication from the application side regarding the dns record set format which is why we had to update the template. So not sure if we should increment the version as its not actually an update on the application side. If you think we should increment the version number regardless, please let me know and I can make that change

wdbaker-godaddy:
Because you are just changing how the variables are passed, I think it's up to your discretion on updating the version number (to distinguish zones added the "old" or "new" way).  If you had added a new record to the template, let's say, then a new version would definitely be warranted.

Thekkedath:
Thank you for explaining that, it makes sense now. I think we will keep the version number as it is now and will increment if new records are added to the template as you suggested.

PR #259, opened 2022-08-12
Increase the pan-resolution of domain names to match subdomians

Swen-Gan:


pawel-kow:
Semantically this template is OK.
Not sure what DNS providers would say about this catch-all wildcard CNAME record. Adding @wdbaker-godaddy and @lpanatau-ionos as reviewers.
Theoretically it would not render any conflict, because any explicit record on a host hides the CNAME. @Swen-Gan please consider if you can live with that limitation (say if the host would be foo.example.com, the A record would stop resolving as soon as for example MX record is placed on foo.example.com).

pawel-kow (approved):
lgtm!

PR #277, opened 2022-10-17
Instantly template created

leopinzon:


goentri.com.instantly.json
```diff
@@ -0,0 +1,26 @@
...
+    {
+      "type": "TXT",
+      "host": "_dmarc",
+      "data": "%dmarcValue%",
+      "ttl": 3600
+    },
+    {
+      "type": "SPFM",
+      "host": "@",
+      "spfRules": "include:%thirdPartyDomain%"
```
pawel-kow, line 23: Generally I think it's OK but I'd love to hear from DNS provider if a variable in spfRules is supported
@wdbaker-godaddy @lpanatau-ionos

abester101: @pawel-kow I think we already do this here on this approved template: https://github.com/Domain-Connect/Templates/blob/master/goentri.com.mailshake.json

pawel-kow (approved):

PR #296, opened 2022-12-06
canva: Remove SSL verification record from canva.com.website.json

garwahl:
This PR removes the `ssl verification` record from the `canva.com.website.json` template.


garwahl:
cc @pawel-kow PTAL when you get a chance

pawel-kow (approved):

PR #297, opened 2022-12-13
Create trady.com.website.json

dommholland:
Adding Trady.com as a service provider to Domain Connect


trady.com.website-subdomain.json
```diff
@@ -0,0 +1,19 @@
...
+   "serviceName":"Trady Subdomain",
+   "version": 1,
+   "syncPubKeyDomain" : "trady.com",
+   "logoUrl":"https://logos.trady.com/logo.png",
+   "description":"Enables a subdomain to work with Trady",
+   "variableDescription": "hostValue is the dynamic host value assigned to the domain",    
+   "records":[  
+      {  
+         "type":"CNAME",
+         "host":"%hostValue%",
```
pawel-kow, line 14: For a subdomain template you'd rather want to use "host" parameter when setting up the template.
Therefore the records would get "@" as a value for "host" member, and the template should get hostRequired=true.
See [spec](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#691-variables-and-hostname-in-template) for details

pawel-kow (changes requested):

pawel-kow (approved):
All good

PR #300, opened 2022-12-30
Add Coda as a Domain Connect Service Provider

johnli-codaio:
Adding Coda as a Domain Connect Service Provider.

The imageUrl is still that of Wix; what is the best way to upload Coda's logo for this?


coda.io.website.json
```diff
@@ -0,0 +1,31 @@
+{  
+    "providerId":"coda.io",
+    "providerName":"Coda",
+    "serviceId":"website",
+    "serviceName":"Coda Site",
+    "version": 1,
+    "syncPubKeyDomain" : "coda.io",
+    "logoUrl":"https://domainconnect.org/wp-content/uploads/2016/06/wix-300x103.png",
```
johnli-codaio, line 8: This is incorrect, but wondering what the process is to get the proper logo uploaded to your CDN?
Will update this PR whenever that is done.

dweitzman-codaio: Looks like these logos can come from any domain. The ones pointed at domainconnect are probably either using a default logo or are some kind of example domain

coda.io.website.json
```diff
@@ -0,0 +1,31 @@
...
+    "description":"Enables a custom domain to work with Coda",
```
dweitzman-codaio, line 9: Could be worthwhile to use a more specific description like, "Allows serving Coda documents from your own domain"

coda.io.website.json
```diff
@@ -0,0 +1,31 @@
...
+       {  
+          "type":"CNAME",
+          "host":"%host%",
+          "pointsTo":"%subdomain%.coda.io",
+          "ttl":3600
+       },
+       {
+           "type": "TXT",
+           "host": "%verification%",
+           "data": "Doc:%docId%",
```
dweitzman-codaio, line 21: It'd be more future-proof to make the verification data opaque here with no mention of docs or doc IDs.

Seems annoying to change or update an existing domain connect template, but in the future we may want to allow domain owners to assert that they're willing to serve any document owned by a particular Coda user ID or any document belonging to a particular Coda workspace

coda.io.website.json
johnli-codaio: Yeah. I'm in contact with the DomainConnect folks in Slack. Looks like there is a process to add the logos, just want to know what that entails.

Screen Shot 2022-12-30 at 5 06 30 PM

coda.io.website.json
johnli-codaio: Sounds like the changing is done via version updating? I know that Domain Connect folks prefer as much static data as possible, and to only really use variables whenever data is not static:

Screen Shot 2022-12-30 at 5 10 40 PM

https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#variables-and-security

pawel-kow:
Hi @johnli-codaio @dweitzman-codaio,
am I right to assume this template is still WIP and not ready to review/merge? If so please add "WIP" to the title and remove when it's ready.

coda.io.website.json
johnli-codaio: Added a logo from Coda's CDN

johnli-codaio:
> 

@pawel-kow I think we're ready now. For things like the `syncPubKey`, we don't have that setup right just now in our DNS settings, is that fine to setup later?

coda.io.website.json
```diff
@@ -0,0 +1,31 @@
...
+    "serviceName":"Coda Site",
+    "version": 1,
+    "syncPubKeyDomain" : "coda.io",
+    "logoUrl":"https://cdn.coda.io/icons/png/color/coda-192.png",
+    "description":"Allows serving Coda documents from your own domain",
+    "variableDescription":"`host` refers to the subdomain that the user wishes to use (not just limited to 'www'). `subdomain` refers to the subdomain for Coda's custom domain server. `docId` refers to the coda document used for custom domain. `verification` refers to the TXT host for doc ownership verification, differs based on A or CNAME. `ipAddress` refers to the proper IP for an apex domain to point to.",
+    "records":[  
+       {  
+          "type":"CNAME",
+          "host":"%host%",
```
pawel-kow, line 14: Please do not use `%host%` as variable as it is a reserved name and typically you don't need to use it in the template.
See [spec Section 6.7](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#hostname-in-template) for details how it works.

I am also not sure how this template should work with both APEX A-Record and a CNAME on subdomain.
Maybe you are looking for 2 different templates for a subdomain and APEX or you need groups to selectively apply only part of the records?

pawel-kow (changes requested):

coda.io.website.json
dweitzman-codaio: I think `host` is supposed to be `_coda.%host%`, which seems like what the linked documentation is talking about

Personally I don't see any added security in forcing `data` TXT contents begin with `Doc`: it won't prevent any hijacking (people can validate the wrong doc while still using a `Doc:` prefix) vs having an opaque blob but it'd makes it harder for us to add new features in the future

coda.io.website.json
```diff
@@ -0,0 +1,31 @@
...
+       {
+           "type": "TXT",
+           "host": "%verification%",
+           "data": "Doc:%docId%",
+           "ttl":3600
+       },
+       {
+           "type": "A",
+           "host": "@",
+           "pointsTo": "%ipAddress",
```
dweitzman-codaio, line 27: Seems like if we want to use three different IP addresses, we may need three variables here? I see other templates doing that sort of thing and I don't see any documentation about variables allowing multiple IP addresses

coda.io.website.json
dweitzman-codaio: Ah thanks, sounds like we probably do want either multiple templates or groupId. Maybe the simple option is two group names like `cname` and `apex`

Just to clarify the intended behavior:

For a subdomain like `www.example.com`:
- Set a `CNAME` record on `www.example.com` pointing to `<variable>.coda.io`. The CNAME `pointsTo` value is variable so that we can use different CNAME in testing environments vs production.
- Set a `TXT` record at `_coda.www.example.com` with a verification blob to prove to Coda the the document owner really controls the subdomain
- _DO NOT_ set an `A` record. Rely entirely on CNAME

For an apex domain like `example.com`:
- Set three `A` records on `example.com` with a few fixed IPs for redundancy
- Set a `TXT` record at `_coda.example.com` with a verification blob to prove to Coda the the document owner really controls the domain
- _DO NOT_ set a `CNAME` (which should be impossible anyway?...)

coda.io.website.json
pawel-kow: @dweitzman-codaio the advisory in the specification to keep the variable data portion as small as possible is mostly related to the templates not using signing, where such template might be used as phishing vector to pass for example DNS validation for SSL certificate issue. With signing enabled the risk is obviously lower but it still should be considered in the template creation.

coda.io.website.json
pawel-kow: Yes, 3 A records with 3 IP variables would be needed, if the IPs are variable all the times.

coda.io.website.json
pawel-kow: I think you'd be better off with 2 templates, like in those templates https://github.com/Domain-Connect/Templates/blob/master/zenfolio.customsubdomain.json and https://github.com/Domain-Connect/Templates/blob/master/zenfolio.customdomain.json.
1 template with groups won't be working for you, as CNAME is not possible on APEX and `hostRequired` would need to be set, on the other side it blocks usage of this template for APEX.

coda.io.apexdomain.json
```diff
@@ -0,0 +1,37 @@
+{  
+    "providerId":"coda.io",
+    "providerName":"Coda",
+    "serviceId":"apexdomain",
+    "serviceName":"Coda Site",
+    "version": 1,
+    "syncPubKeyDomain" : "coda.io",
+    "logoUrl":"https://cdn.coda.io/icons/png/color/coda-192.png",
+    "description":"Allows serving Coda documents from your own domain (apex only).",
+    "variableDescription":"`ipAddress[1, 2, 3]` refers to the IP addresses of Coda's custom domain server. `docId` refers to the coda document. `environment` refers to the environment for Coda's custom domain server. `docId` refers to the coda document used for custom domain. `verification` refers to the TXT verification of doc ownership",
```
dweitzman-codaio, line 10: The description looks out-of-date and has some repetition. Seems like `docId` is just `verification` now.

coda.io.apexdomain.json
```diff
@@ -0,0 +1,37 @@
+{  
+    "providerId":"coda.io",
+    "providerName":"Coda",
+    "serviceId":"apexdomain",
+    "serviceName":"Coda Site",
+    "version": 1,
+    "syncPubKeyDomain" : "coda.io",
```
dweitzman-codaio, line 7: Looks like some templates use a dedicated subdomain named something like `domainconnect.coda.io` for public keys. Might be a good idea for us to use that pattern too, here and in the other template.

If I understand correctly, the keys themselves would live in TXT records at a subdomain like `<keyId>.domainconnect.coda.io`

dweitzman-codaio (commented):
Looks reasonable to me.

Might be worthwhile to lower the TTLs to something more like 300 (or 600)?

Seems like a lot of people struggle with DNS configuration and my guess it people won't care at all if there's an extra roundtrip for DNS lookup once every 5 or 10 minutes, but they probably will care if they have to wait an hour for setting changes to propagate

coda.io.apexdomain.json
pawel-kow: yes

coda.io.subdomain.json
```diff
@@ -0,0 +1,25 @@
...
+    "records":[  
+       {  
+          "type":"CNAME",
+          "host":"%subdomain%",
+          "pointsTo":"%environment%.coda.io",
+          "ttl":3600
+       },
+       {
+           "type": "TXT",
+           "host": "_coda.%subdomain%",
```
pawel-kow, line 20: Same here. It's just `_coda`

coda.io.subdomain.json
```diff
@@ -0,0 +1,25 @@
...
+    "serviceName":"Coda Site",
+    "version": 1,
+    "syncPubKeyDomain" : "coda.io",
+    "logoUrl":"https://cdn.coda.io/icons/png/color/coda-192.png",
+    "description":"Allows serving Coda documents from your own domain (subdomain only)",
+    "variableDescription":"`subdomain` refers to the subdomain that the user wishes to use (not just limited to 'www'). `environment` refers to the environment for Coda's custom domain server. `docId` refers to the coda document used for custom domain. `verification` refers to the TXT verification of doc ownership",
+    "records":[  
+       {  
+          "type":"CNAME",
+          "host":"%subdomain%",
```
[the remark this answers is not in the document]
johnli-codaio: Looking at the spec, it looks like host is used for a static string. What if the host is up to the discretion of the user?

for ex: Say someone wants to setup either a custom domain `www.example.com` as well as `abc.example.com`

We at Coda would like to be able to support someone's ability to choose whatever host they choose, rather than sticking to just `www`.

I'm looking at the extenuating circumstances for adding variables for the `host` parameter. This `host` is only really set for TXT and for CNAME since we wanted to respect any host for CNAMEs only (A record we only use for APEX publishing, so there is no host). If this is the case, should I just leave `subdomain` and pass that as a query? Our TXT record also should respect the custom host that a user passes in, which is why I have it as `_coda.%subdomain%`. Is there another way to honor this requirement for the template?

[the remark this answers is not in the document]
pawel-kow: @johnli-codaio if you place the configuration `"host": "@"` in CNAME and `"host"="_coda"` for TXT and then when applying the template you'd use `host=www` or 'host=abc` query parameter, the DNS provider will create records with the desired subdomain. This is the purpose of this reserved `host` parameter to do exactly that.
If you mimic the same with own variable, you will run into trouble if you ever try to add the templates to more than one subdomain of the same domain. Some DNS providers take domain+host pair as unique instance key, so any successive apply will be treated as an update and other records (you other subdomain) deleted.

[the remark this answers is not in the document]
johnli-codaio: AHHHH I see, I didn't make the connection that `host` can be passed as a query parameter and that automatically hooks up things. When I saw parameter, I assumed it was a template parameter, instead of a query parameter passed in.

That's perfect, then I'll do as you say! Thanks again for clarification :).

johnli-codaio:
@pawel-kow this should be ready for a re-review!

coda.io.subdomain.json
```diff
@@ -0,0 +1,25 @@
+{  
+    "providerId":"coda.io",
+    "providerName":"Coda",
+    "serviceId":"subdomain",
+    "serviceName":"Coda Site",
+    "version": 1,
+    "syncPubKeyDomain" : "domainconnect.coda.io",
```
pawel-kow, line 7: One last change I mentioned before, this template requires `"hostRequired": true`

pawel-kow (changes requested):

pawel-kow (approved):

PR #306, opened 2023-01-20
Create goentri.com.fairmint.json

evelyndearaujo-goentri:


goentri.com.fairmint.json
```diff
@@ -0,0 +1,45 @@
...
+    "serviceId": "fairmint",
+    "serviceName": "Fairmint",
+    "version": 1,
+    "logoUrl": "https://cdn.goentri.com/logo.svg",
+    "description": "Allows user to easily set up domain using Entri",
+    "variableDescription": "webSitePointsTo1, webSitePointsTo2, webSitePointsTo3, webSitePointsTo4 and webSitePointsTo5 is to where the website ins pointed to and CnameHost1, CnameHost2,CnameHost3, CnameHost4, CnameHost5 is the host for each CNAME",
+    "syncPubKeyDomain": "goentri.com",
+    "syncRedirectDomain": "api.goentri.com, goentri.com, entri.com, app.entri.com,app.goentri.com",
+    "sharedProviderName": true,
+    "records": [
```
pawel-kow, line 13: This is rather a general question related to this template. Would it be really adding all 5 CNAMEs each time (as there are no groups defined) and are all those CNAMEs always dynamic on both host and target side?
Such templates create typical difficulties for the DNS providers in conflict management etc. therefore should be used sparingly if the use-case really justifies it.

leopinzon: Hey Pawel! thanks for you pointing this out, we spoke with the customer and they're still in the hosts/values definition process. We will start with the 5 cnames as variables and will fine tune the template once they have the final setup.

pawel-kow: ok, if this a testing stage, would you maybe name the serviceId with -test or -dev suffix and maybe the description, so that it's more clear for the DNS providers?

evelyndearaujo-goentri: Hi Pawel.
Sure.
I just updated the serviceId and the description. Thank you for your feedback.

pawel-kow (approved):
LGTM

leopinzon:
Thanks a lot @pawel-kow

PR #309, opened 2023-01-25
Revert "Add spf record for GoDaddy and Whitelabel Managed WordPress"

jbardo-godaddy:
Reverts Domain-Connect/Templates#292


olafleur-godaddy:
@wdbaker-godaddy can we merge?

PR #311, opened 2023-01-27
Create constantcontact.com.email.json

adancus:


wdbaker-godaddy (commented):
I see that you have 'priority' included with these 2 CNAME records.  Please remove that attribute if indeed these are to remain as CNAME records.

wdbaker-godaddy (approved):
Approved

PR #319, opened 2023-02-01
Entri/mighty sites template

evelyndearaujo-goentri:


wdbaker-godaddy (approved):
Approved

goentri.com.fairmint.json
```diff
@@ -0,0 +1,45 @@
+{
```
pawel-kow, line 1: This file is in #306, so should not be here, right?

goentri.com.mighty-sites.json
```diff
@@ -0,0 +1,27 @@
...
+  "serviceId": "mighty-sites",
+  "serviceName": "MightySites",
+  "version": 1,
+  "logoUrl": "https://cdn.goentri.com/logo.svg",
+  "description": "Allows user to easily set up domain using Entri",
+  "syncPubKeyDomain": "goentri.com",
+  "syncRedirectDomain": "api.goentri.com, goentri.com, entri.com, app.entri.com,app.goentri.com",
+  "variableDescription": "mappingValue1, mappingValue2 are the values assigned to hosts websiteHost1, websiteHost2 subsequently",
+  "sharedProviderName": true,
+  "records": [
```
pawel-kow, line 13: A similar comment like in #306, are those values really truly dynamic in all cases?
For me it looks like a typical APEX/www configuration, which by the way may have flaws as there might be some implementation differences if you pass `@` as a value for `%websiteHost1%`

pawel-kow (changes requested):

goentri.com.fairmint.json
evelyndearaujo-goentri: Yes, you are right, it's deleted. Thank you.

pawel-kow (approved):
Lgtm

PR #320, opened 2023-02-06
Update goentri.com.mighty-sites.json

evelyndearaujo-goentri:
deleted hostRequired which means that now it's set to false


pawel-kow (approved):
LGTM

PR #321, opened 2023-02-13
Remove the sync block on hubspot-template-3

chintans1:
We're looking to start using this template for the Domain Connect sync flow, therefore we're removing the `syncBlock` on it. For our sync requests, we will digitally sign the requests to still maintain the security aspect of the flow.

### DIFF
--- a/hubspot.com.hubspot-template-3.json
+++ b/hubspot.com.hubspot-template-3.json
@@ -3,6 +3,8 @@
     "providerName": "HubSpot",
     "serviceId": "hubspot-template-3",
     "serviceName": "HubSpot CMS Site",
+    "syncPubKeyDomain": "domainconnect.hubspot.com",
+    "syncRedirectDomain": "app.hubspot.com, app.hubspotqa.com, api.hubspot.com, api.hubspotqa.com",
     "version": 1,
     "logoUrl": "https://domainconnect.org/wp-content/uploads/2017/02/HubSpot-768x223.png",
     "description": "Enables an apex domain to work with the HubSpot CMS",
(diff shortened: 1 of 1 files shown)


hubspot.com.hubspot-template-3.json
```diff
@@ -4,7 +4,6 @@
     "serviceId": "hubspot-template-3",
     "serviceName": "HubSpot CMS Site",
     "version": 1,
-    "syncBlock": true,
```
pawel-kow, line 7: The template in this form would not be secure if `syncBlock` is removed.
If you intend to use signing, please add  `syncPubKeyDomain` in the same PR.
You may also want to add `syncRedirectDomain` for the sync flow, if you plan to use redirects.

pawel-kow (changes requested):

hubspot.com.hubspot-template-3.json
chintans1: Thanks Pawel! I just updated it so this should be good now. Let me know if you have any more concerns!

hubspot.com.hubspot-template-3.json
```diff
@@ -3,8 +3,9 @@
     "providerName": "HubSpot",
     "serviceId": "hubspot-template-3",
     "serviceName": "HubSpot CMS Site",
+    "syncPubKeyDomain": "domainconnect.hubspot.com",
+    "syncRedirectDomain": "app.hubspot.com, app.hubspotqa.com, api.hubspot.com, api.hubspotqa.com",
     "version": 1,
```
pawel-kow, line 8: One more thing, please bump the version. Sorry I did not notice that before.

pawel-kow (changes requested):

hubspot.com.hubspot-template-3.json
chintans1: Done :)

pawel-kow (approved):
Lgtm!

PR #325, opened 2023-02-17
Update hubspot-cname-txt template for new TXT value

chintans1:
We are looking to use this for a sync flow as well in the future so I added the property `syncRedirectDomain`.


wdbaker-godaddy (approved):
Approved

PR #334, opened 2023-03-02
Updated TTL for jamfeed

smendivil:


pawel-kow (changes requested):
If the template was deployed already by any DNS provider consider bumping the version number.

pawel-kow (approved):

PR #342, opened 2023-03-17
Add syncRedirectDomain to hubspot-cname and hubspot-txt template

chintans1:
Also bumped the versions to `2` for this change. We're trying to utilize these templates as well in a sync flow.


pawel-kow (approved):
lgtm!

PR #346, opened 2023-03-21
Update the ip address variables to have an appending % for Coda ApexDomain json

johnli-codaio:
I realized the variable we had was syntactically incorrect. Just fixing that up so GoDaddy can properly onboard with us.


pawel-kow (approved):
lgtm!

PR #354, opened 2023-03-31
Update goentri.com.squareup.json

evelyndearaujo-goentri:


goentri.com.squareup.json
```diff
@@ -3,25 +3,26 @@
...
+        "ttl": 600
       },
       {
-        "type":"A",
-        "host":"www",
-        "pointsTo":"%ip%",
-        "ttl": 300
+         "type": "CNAME",
+         "host": "www",
+         "pointsTo": "%domain%",
```
pawel-kow, line 24: Hi, you'd rather want to use `@` instead of `%domain%`, otherwise if the template is used with a subdomain for example `host=foo` it will generate `www.foo.example.com` pointing to `example.com` instead of `foo.example.com`.

pawel-kow (changes requested):

goentri.com.squareup.json
```diff
@@ -3,25 +3,26 @@
...
     "serviceName": "Squareup",
-    "version": 1,
+    "version": 2,
     "logoUrl": "https://cdn.goentri.com/logo.svg",
     "description": "Allows user to easily set up domain using Entri",
     "syncPubKeyDomain": "goentri.com",
     "syncRedirectDomain": "api.goentri.com, goentri.com, entri.com, app.entri.com,app.goentri.com",
-    "variableDescription":"ip is the ip to point the domain",
+    "variableDescription":"ip is the ip to point the domain, aHost is whe A record is pointed, cnameHost is the host for the CNAME record and cnamePointsTo is where the CNAME is pointed ",
+    "hostRequired": true,
```
pawel-kow, line 12: I'm really confused by this template.
If you set hostRequired it means you expect always to set the host variable and generate a subdomain.
Then `%aHost%` and `%cnameHost%` will generate a subhost of this subdomain each. Sure about it?
The v1 of the template looked like a typical main domains + www subdomain setup.

evelyndearaujo-goentri: Hey Pawel, I've made new changes to avoid the behavior you reported above.

evelyndearaujo-goentri: Hi Pawel,

I've made some changes to avoid this behavior you reported above.

Thank you.

pawel-kow (approved):
Lgtm

PR #365, opened 2023-04-18
Create Service Provider template for Seguno

mbaumbach:
This is the service provider template for [Seguno Software](https://www.seguno.com). We intend to provide a mechanism to insert DNS records for DKIM domain authentication of our customers.

### DIFF
--- a/seguno.com.domain-authentication.json
+++ b/seguno.com.domain-authentication.json
@@ -6,7 +6,7 @@
   "version": 1,
   "syncPubKeyDomain": "seguno.com",
   "logoUrl": "https://cdn.seguno.com/images/seguno-logo.png",
-  "description": "Places SPF and DKIM records to authenticate email sent by Seguno Software on behalf of the user",
+  "description": "Places Return-Path/bounce domain and DKIM records to authenticate email sent by Seguno Software on behalf of the user",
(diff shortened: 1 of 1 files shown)


seguno.com.domain-authentication.json
```diff
@@ -0,0 +1,25 @@
+{
+  "providerId": "Seguno.com",
+  "providerName": "Seguno",
+  "serviceId": "domain-authentication",
+  "serviceName": "Seguno Domain Authentication",
+  "version": 1,
+  "syncPubKeyDomain": "seguno.com",
+  "logoUrl": "https://cdn.seguno.com/images/seguno-logo.png",
+  "description": "Places SPF and DKIM records to authenticate email sent by Seguno Software on behalf of the user",
```
pawel-kow, line 9: for SPF I would have expected SPFM record. So either the description is wrong or the record is missing

pawel-kow (changes requested):

seguno.com.domain-authentication.json
mbaumbach: Thanks @pawel-kow, this was an incorrect description. It's not an SPF record but one used for [Return-Path/bounce domain](https://support.sparkpost.com/docs/tech-resources/custom-bounce-domain). We use Sparkpost as a sending platform and the SPF is handled automatically.

PR #370, opened 2023-05-10
Create goentri.com.realnice.json

evelyndearaujo-goentri:
Hi team, Nice Site is changing its name to Real Nice and requires us to update its template.


pawel-kow (approved):
lgtm!

PR #382, opened 2023-06-02
Create AcyMailer Template

roumilb:
Template for the AcyMailer service. It's our service to send emails.
On the setup, users need to create CNAME records with randomly generated CNAME records


acymailing.com.acymailer.json
```diff
@@ -0,0 +1,34 @@
...
+    "serviceId": "acymailer",
+    "serviceName": "AcyMailer DKIM Authentication",
+    "version": 1,
+    "logoUrl": "https://www.acymailing.com/wp-content/uploads/2019/05/ACYMAILING_logo_mobile.png",
+    "description": "Allows user to easily set up CNAME record for AcyMailer service",
+    "syncPubKeyDomain": "acymailer.com",
+    "syncRedirectDomain": "api.acymailer.com, acymailer.com",
+    "variableDescription": "mappingValue1, mappingValue2 and mappingValue3 are the values assigned to host websiteValue1, websiteValue2 and websiteValue3",
+    "sharedProviderName": true,
+    "syncBlock": true,
```
pawel-kow, line 13: Sure about it? It will imply only async protocol which is not supported by every DNS provider and more problematic onboarding. Also there are no groups, so obviously no multi-stage configuration which requires async process.

acymailing.com.acymailer.json
```diff
@@ -0,0 +1,34 @@
...
+    "records": [
+        {
+            "type": "CNAME",
+            "host": "%websiteValue1%",
```
pawel-kow, line 17: if this is for DKIM it would be better to have hostnames like "%dkimhost1%._domainkey" to limit the scope. Fully open and generic CNAMEs pose always a security risk so better avoid them if not necessary.

pawel-kow (changes requested):

acymailing.com.acymailer.json
roumilb: Indeed I wanted a synchronous protocol but in the common feedback and the documentation it's written that if we don't specify it, then the `synchronous protocol must be disabled for this template`.  
I've removed it.

acymailing.com.acymailer.json
roumilb: I've changed it to include the `._domainkey` and I've also changed the `pointsTo` to only have the value

PR #383, opened 2023-06-02
Create goentri.com.toast.json

evelyndearaujo-goentri:
Taost has a use case in which the user can add until 10 subdomains besides the standard configuration, this is the reason I created 12 groups separately.

### DIFF
--- a/goentri.com.toast.json
+++ b/goentri.com.toast.json
@@ -3,7 +3,7 @@
   "providerName": "Entri",
   "serviceId": "toast",
   "serviceName": "Toast",
-  "version": 1,
+  "version": 2,
   "logoUrl": "https://cdn.goentri.com/logo.svg",
   "description": "Allows user to easily set up domain using Entri",
   "syncPubKeyDomain": "goentri.com",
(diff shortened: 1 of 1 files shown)


pawel-kow (commented):
Technically the template is OK, however I must ask the question, why so much of redundancy?
You know that it's possible to apply more than one group at once, don't you?
This means, that each subdomain could be just a separate group with one record and then you'd only need to apply the groups as you see fit, having like ca ~15 records in the template instead of 89.

pawel-kow:
@evelyndearaujo-goentri it's not possible (yet?) to repeat the same group several times with one an the same apply flow. But you can place one subdomain per group s1,s2,... und then apply several groups with appropriate variables. The testing tool shows you how it works by selecting several checkboxes.

evelyndearaujo-goentri:
Are you suggesting creating 10 groups only with the CNAMEs for the subdomain and letting the standard records (A,TXT and CNAME) in a separate group? Am I getting it right?

leopinzon:
This is pure gold @pawel-kow, thank you

evelyndearaujo-goentri:
@pawel-kow I've made some updates and also added some TXT required by them

pawel-kow (approved):
Ok...

PR #401, opened 2023-07-07
Update goentri.com.toast.json

evelyndearaujo-goentri:


pawel-kow (changes requested):
version number bump

pawel-kow (commented):
No need to update version?

evelyndearaujo-goentri:
I prefer to keep version 2 considering the last template was not correct. Thanks for checking...

evelyndearaujo-goentri:
actually, you're right, let me update it on our side.

pawel-kow (approved):
thx

PR #406, opened 2023-07-13
Update goentri.com.mixo.json

evelyndearaujo-goentri:


pawel-kow (changes requested):
version bump?

evelyndearaujo-goentri:
> version bump?

it was updated to version 3, is that right?

pawel-kow (approved):
lgtm

PR #407, opened 2023-07-14
Update goentri.com.clickmax.json

evelyndearaujo-goentri:


goentri.com.clickmax.json
```diff
@@ -17,6 +17,11 @@
       "host": "@",
       "pointsTo": "%cnamePointsTo%",
       "ttl": 3600
+    },
+    {
+      "type": "REDIR301",
+      "target": "%target%",
+      "host": "@"
```
pawel-kow, line 24: no a chance this template would work. CNAME cannot co-exist with A/AAAA records needed for the redirect on the same host.

pawel-kow (changes requested):

goentri.com.clickmax.json
evelyndearaujo-goentri: Would this work if the host required is set to false and the CNAME host is back to www?

pawel-kow: yes, if you place CNAME on "www" then host required can be set to false

evelyndearaujo-goentri: it's done.

pawel-kow (approved):
lgtm

PR #410, opened 2023-07-24
feat: Added Combo Records

b805rohit:
feat: Combo Records


pawel-kow (changes requested):
The changes look OK.
Some templates have groups on a single records, which might not be needed.

Generally - the updated templates should get a new higher version.

pawel-kow (approved):
lgtm

PR #411, opened 2023-07-26
Entri templates

evelyndearaujo-goentri:


pawel-kow (approved):
lgtm!

@evelyndearaujo-goentri did you know you can also refer to the same domain in the redirect without using any external variable, just `%fqdn%` ?
```
    {
      "type": "REDIR301",
      "target": "https://www.%fqdn%",
      "host": "@",
      "groupId": "a2"
    },
```

evelyndearaujo-goentri:
> 

I didn't know that, thanks for sharing it @pawel-kow. In the case, the"fqdn" is the root domain, is that right?

pawel-kow:
`%fqdn%` is the full hostname where the template is applied.
If there is no `host` parameter specified, this is the domain name, if `host` is specified than it is the full name of the subdomain.

evelyndearaujo-goentri:
got it, thanks. 

Do you prefer I update the template using the %fqdn% ?

PR #412, opened 2023-07-27
Update goentri.com.sendmarc.json

evelyndearaujo-goentri:


goentri.com.sendmarc.json
```diff
@@ -23,7 +23,7 @@
       "type": "SPFM",
       "host": "@",
       "groupId": "a2",
-      "spfRules": "include:%thirdPartyDomain%"
+      "spfRules": "redirect=%thirdPartyDomain%"
```
pawel-kow, line 26: redirect rules are not currently supported by SPFM, as those cannot be reliably merged with the other rules

pawel-kow (changes requested):

goentri.com.sendmarc.json
evelyndearaujo-goentri: Hi @pawel-kow, thank you for warning me about that.

What would be the right configuration for the following SPF?

{
        "value": "v=spf1 redirect=_sokiruvgn.sdmarc.net",
        "host": "@",
        "ttl": 3600,
        "type": "TXT"
      }

pawel-kow: Sth like:
```
    {
      "data": "v=spf1 redirect=%thirdPartyDomain%",
      "host": "@",
      "ttl": 3600,
      "type": "TXT",
      "txtConflictMatchingMode": "Prefix",
      "txtConflictMatchingPrefix": "v=spf1",
      "essential": "OnApply"
    }
```

evelyndearaujo-goentri: thank you

pawel-kow: ... and bump the version number

pawel-kow (approved):
lgtm

goentri.com.sendmarc.json
evelyndearaujo-goentri: @pawel-kow considering Google Domains don't support "txtConflictMatchingMode": "Prefix","txtConflictMatchingPrefix": "v=spf1", "essential": "OnApply

Is there a different way to configure that record?

pawel-kow: txtConflictMatchingMode, txtConflictMatchingPrefix and essential have a semantic meaning which will make this alternative method secure to use.
Without txtConflictMatching* setting it will be possible to set up 2nd SPF, which will basically void it.
"essential" setting is only important for providers tracking the template usage and removing whole templates. So Google likely can tell if they are affected. If this setting is not there it would mean the template dropped if SPF is modified or deleted. Normally not a wanted behavior.

Functionally Google could onboard the template without these parameters and it would work for most of the cases.

Alternative is to convince the provider to use include instead of redirect, and keep using SPFM.

pawel-kow: txtConflictMatchingMode, txtConflictMatchingPrefix and essential have a semantic meaning which will make this alternative method secure to use.
Without txtConflictMatching* setting it will be possible to set up 2nd SPF, which will basically void it.
"essential" setting is only important for providers tracking the template usage and removing whole templates. So Google likely can tell if they are affected. If this setting is not there it would mean the template dropped if SPF is modified or deleted. Normally not a wanted behavior.

Functionally Google could onboard the template without these parameters and it would work for most of the cases.

Alternative is to convince the provider to use include instead of redirect, and keep using SPFM.

PR #433, opened 2023-09-06
contribute domain-connect-template-linter tool

kerolasa:
This tool tries to make it easy to check and see if a template is valid. The tool has Cloudflare speficic -cloudflare option.  This option is meant for service providers who want to onboard to Cloudflare.  Other DNS providers are welcome to add their own hints about settings.

The basic validation attempts to follow the Domain Connect Spec as close as possible.  When validation issues happen output is provided to stderr.

Finally there is -pretty option that will unify template json format.  This is technically irrelevant, but it is nice for humans to have some regularity in the template files.

Reference: https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#52-template-definition


pawel-kow:
Hi @kerolasa, thanks for the contribution! Just thinking, should it sit inside Template repo or maybe it would be better having own repo for this project? It would be imho easier to maintain without the process we follow in the templates repo. What do you think?
Also, can this linter do more checks than the current json schema? If yes it would be interesting to integrate it into the CI pipeline on github.

kerolasa:
> should it sit inside Template repo or maybe it would be better having own repo

There are pros and cons in both ways.  Dedicated repo for a template
validation tool does make data to be more hygienic.  But having it in
separate repo also makes it harder for common user to have the tool readily
available when dealing with the templates.

That said I have a hunch the templates are primarily changed with the github
web UI.  In that scenario additional command-line tooling will never be part
of user work flow. 

I must admit I haven't looked into github ci, so I cannot tell how hard
would it be to do that integration.  Especially if/when web browser based
file edits need to be covered.  Assuming integration is doable having
automatic checks would be excellent.

> can this linter do more checks than the current json schema?

Code is code.  It's possible to make it to what is needed.  I suppose the
primary question with improvements like this is how thing ought to be?  IMHO
using github issues / pull requests to these discussions is the natural
thing to do.  Essentially I am proposing 'happy go lucky, lets see how
things develop' process.

kerolasa:
FYI more checks was  added in commit 8f791eab06d2bf92ef121dafce74627b9464e94a

kerolasa:
Couple more improvements. When stderr is not terminal colors are suppressed from output. Mixing CNAME with other records is attempted to catch, I'm sure with variables one can still shoot oneself to foot. And -logos option is added, when defined it will check the logo urls return http/200. The -logos is off by default because it 1) requires network 2) is slow (when timing out).

pawel-kow:
> > should it sit inside Template repo or maybe it would be better having own repo
> 
> There are pros and cons in both ways. Dedicated repo for a template validation tool does make data to be more hygienic. 

Yes, I would tend to this solution. Also the management would be easier. I'm happy to create a project under Domain-Connect org, so that it is in the same place and make a reference in the README.
It would be also easier to integrate into github CI.

> But having it in separate repo also makes it harder for common user to have the tool readily available when dealing with the templates.

I guess if someone is able to come through a hurdle of compiling it, it would not make too much difference.

> That said I have a hunch the templates are primarily changed with the github web UI. In that scenario additional command-line tooling will never be part of user work flow.

Yes, that's why adding it to CI would be relevant.

> I must admit I haven't looked into github ci, so I cannot tell how hard would it be to do that integration. Especially if/when web browser based file edits need to be covered. Assuming integration is doable having automatic checks would be excellent.

In-editor linting is not possible in github AFAIK, but CI would at least show an error when running through the PR pipeline. This works with schema check already - no matter if someone makes changes on a side or in github editor.

kerolasa:
OK, lets add new repo. Feel free to rename the command. The current `domain-connect-template-linter` feels a bit long. That said I don't have good ideas how to name this. As the joke goes there are only two hard problems in programming: naming things, cache invalidation and off by one errors.

pawel-kow:
I created https://github.com/Domain-Connect/dc-template-linter and added you as admin.

pawel-kow:
I moved the code and edited README, but feel free to override with your own commit history if you wish.

kerolasa:
With code in https://github.com/Domain-Connect/dc-template-linter there is just need for cleaning up README file example in this repository. I opened new pull request about that: https://github.com/Domain-Connect/Templates/pull/437

PR #436, opened 2023-09-14
Update getreadydigital.com.cname-record.json

mike-kirby-dev:
Following review of the Template Record section of the spec, this change reflects a fix based on the Description for Type CNAME to change the host to a variable from '@' - "CNAME: host, pointsTo, TTL (host must not be null or @)" (https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#53-template-record). This is also highlighted as an error in the template linter tool for our template.

### DIFF
--- a/getreadydigital.com.cname-record.json
+++ b/getreadydigital.com.cname-record.json
@@ -7,14 +7,14 @@
   "version": 1,
   "logoUrl": "https://getreadydigital.com/wp-content/uploads/2023/04/cropped-GR-Logo-Favicon-192x192.png",
   "description": "Enable root domain to work with the GetReadyDigital",
-  "variableDescription": "CNAMEs: GetReadyDigital connection",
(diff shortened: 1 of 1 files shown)


pawel-kow:
Template is OK. With `hostRequired` set to `true` such template is fully functional and can be used.
Good catch with the spec. I found the spot, where the erroneous or inconsistent text remained.
I created a PR to clarify this issue https://github.com/Domain-Connect/spec/pull/73
@kerolasa I think you may want to take a look into this as well

kerolasa:
This change looks good to me. And the linter got fix about host check, see the change below.

https://github.com/Domain-Connect/dc-template-linter/commit/bcbaea70fe1a04bbb4755d4078a132b03d51bcd2

pawel-kow (changes requested):
@mike-kirby-dev do you still consider this template change necessary, after the issue with the linter has been corrected?

PR #438, opened 2023-09-20
Change pubkeydomain value

avag89:


easydmarc.com.setup.json
```diff
@@ -6,7 +6,7 @@
   "version": 1,
```
pawel-kow, line 6: Hi, do you want to bump the version, or it was not yet deployed by anyone?

pawel-kow (changes requested):

easydmarc.com.setup.json
avag89: not deployed yet. So it is ok with the current version.

pawel-kow (approved):

PR #439, opened 2023-09-20
Entri/kajabi template

evelyndearaujo-goentri:


goentri.com.kajabi.json
```diff
@@ -0,0 +1,79 @@
...
+      "type": "MX",
+      "host": "kjbm",
+      "pointsTo": "mxa.mailgun.org",
+      "priority": 10,
+      "groupId": "a1",
+      "ttl": 3600
+    },
+    {
+      "type": "MX",
+      "host": "mxb.mailgun.org",
```
pawel-kow, line 52: I don't think this would be correct...

pawel-kow (changes requested):

goentri.com.kajabi.json
evelyndearaujo-goentri: Thank you @pawel-kow, it's fixed.

pawel-kow (approved):
lgtm

PR #441, opened 2023-09-25
Create goentri.com.stunning.json

evelyndearaujo-goentri:


goentri.com.stunning.json
```diff
@@ -0,0 +1,22 @@
...
+  "description": "Allows user to easily set up dkim record using Entri",
+  "syncPubKeyDomain": "goentri.com",
+  "syncRedirectDomain": "api.goentri.com, goentri.com, entri.com, app.entri.com,app.goentri.com",
+   "variableDescription": "cnamePointsTo is the URL to where the subdomain is pointed and cnameHost is the host ",
+  "sharedProviderName": true,
+  "hostRequired": false,
+  "records": [
+    {
+      "type": "CNAME",
+      "host": "%cnameHost%",
```
pawel-kow, line 17: why not `@` host and work with `host` parameter?
Yes, in the end you have the same record created, but some providers (IONOS, GoDaddy AFAIK) are identifying unique templates based on quadruple (domain, host, providerId, serviceId). In this case if you apply a second template with just different `%cnameHost%` the first one will be just removed.

pawel-kow (changes requested):

goentri.com.stunning.json
evelyndearaujo-goentri: Yeah, makes sense. I would also need to change this template anyway, I'll make all those changes soon.

evelyndearaujo-goentri: Thanks

pawel-kow: So you want to keep this a1 group with CNAME here?

evelyndearaujo-goentri: qq - in case they send both a Record and CNAME would we be able to use 2 different templates for the same request?

pawel-kow: 2 templates in the same request are not possible in the protocol at the moment.
If you want to have both in the same request than this template is fine, with a downside of just one instance per domain. As you are setting A record on APEX I believe it's fine like this.

evelyndearaujo-goentri: Thanks @pawel-kow, I'm going to keep the template as it is.

pawel-kow (approved):
lgtm

PR #445, opened 2023-10-03
Create informaten.com.hosting_web.json

Luxaaa:


informaten.com.hosting_web.json
```diff
@@ -0,0 +1,32 @@
+{
+  "providerId": "informaten.com",
+  "providerName": "Informaten",
+  "serviceId": "hosting_web",
+  "serviceName": "Webhosting on Informaten.com",
+  "version": 1,
+  "warnPhishing": true,
```
pawel-kow, line 7: sure about using warn phishing instead of signing the calls?

pawel-kow (changes requested):

Luxaaa:
Yes, We will optimize this later.

Gesendet von Outlook für Android<https://aka.ms/AAb9ysg>
________________________________

pawel-kow:
Well, there is work behind each review on the public repo as well as template deploy at each provider, so better skipping such intermediate version if you plan to go for signing anyway. Note some providers may even not take an insecure template at all.

PR #451, opened 2023-10-20
Create mailchimp.com.email-signing.json

adriankappel:
For proper email signing on a user's behalf, we prompt them to add specific records to the DNS. The hope is by using this protocol, we can perform more of the heavy lifting on their behalf.

It's grouped in two parts:
- `dkim` configuration
- `dmarc` configuration

By checking beforehand, we can determine whether the user needs both groups, or just DKIM (because a DMARC record is already present).

Any and all feedback welcome. Thanks.

### DIFF
--- a/mailchimp.com.email-signing.json
+++ b/mailchimp.com.email-signing.json
@@ -1,0 +1,34 @@
+{
+    "providerId": "mailchimp.com",
+    "providerName": "Intuit Mailchimp",
+    "serviceId": "email-signing",
+    "serviceName": "Intuit Mailchimp Email Signing",
+    "version": "1",
+    "logoUrl": "https://cdn-images.mailchimp.com/product/brand_assets/logos/mc-freddie-dark.svg",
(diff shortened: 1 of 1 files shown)


mailchimp.com.email-signing.json
```diff
@@ -0,0 +1,34 @@
...
+            "groupId": "dkim",
+            "type": "CNAME",
+            "host": "%sel2%._domainkey",
+            "pointsTo": "%host2%",
+            "ttl": 3600
+        },
+        {
+            "groupId": "dmarc",
+            "type": "TXT",
+            "host": "_dmarc",
```
pawel-kow, line 29: On DMARC you may want to set txtConflictMatchingMode = All, so there will be no more than one TXT record on the `_dmarc` host. Not all providers support it, but it's helpful for those who do.

pawel-kow (changes requested):

pawel-kow (approved):
lgtm

PR #458, opened 2023-11-06
Create megapowerani.com.template.json

hackforce-squad:
Freshsales Domain Verification using Domain Connect


megapowerani.com.template.json
```diff
@@ -0,0 +1,70 @@
...
+    "providerName": "Freshsales Test",
+    "serviceId": "template",
+    "serviceName": "Domain verification",
+    "version": 1,
+    "logoUrl": "https://avatars.slack-edge.com/2021-08-02/2349087959585_442632783598439baea4_512.png",
+    "description": "Freshsales Domain Verification Test",
+    "variableDescription": "This is  atest for Freshsales domain verification using Domain Connect",
+    "syncPubKeyDomain": "megapowerani.com",
+    "syncRedirectDomain": "megapowerani.com",
+    "warnPhishing": true,
```
pawel-kow, line 12: Do I oversee something or there is no phishing vector from within this template?
There are no variables and signing in place so this template should be secure AFAIK.

megapowerani.com.template.json
```diff
@@ -0,0 +1,70 @@
+{
+    "providerId": "megapowerani.com",
+    "providerName": "Freshsales Test",
+    "serviceId": "template",
+    "serviceName": "Domain verification",
+    "version": 1,
+    "logoUrl": "https://avatars.slack-edge.com/2021-08-02/2349087959585_442632783598439baea4_512.png",
+    "description": "Freshsales Domain Verification Test",
+    "variableDescription": "This is  atest for Freshsales domain verification using Domain Connect",
```
pawel-kow, line 9: there are no variables, so `variableDescription` seems to be confusing. Likely this should be in `description` instead

megapowerani.com.template.json
```diff
@@ -0,0 +1,70 @@
+{
+    "providerId": "megapowerani.com",
+    "providerName": "Freshsales Test",
```
pawel-kow, line 3: `Test` in Provider name seems to be confusing. 
If this is a test template I would rather see it in `serviceId` / `serviceDescription`.
Note that most of DNS providers would display this name in their UI.

megapowerani.com.template.json
```diff
@@ -0,0 +1,70 @@
...
+    "serviceId": "template",
+    "serviceName": "Domain verification",
```
pawel-kow, line 5: Can you sync `serviceId` with `serviceName`? "template" seems to be very generic and "verification" already quite specific.

pawel-kow (changes requested):

megapowerani.com.template.json
hackforce-squad: @pawel-kow Apologies, I will perform the changes where I mention "Test" in the serviceID/serviceDescription. I am performing this test so that the team who will work on the production application can create an new production template later. I am only preparing this for a demo so that this gets accepted for production.

megapowerani.com.template.json
hackforce-squad: I will have this updated._

megapowerani.com.template.json
hackforce-squad: I will leave variableDescription as empty

megapowerani.com.template.json
hackforce-squad: Yes, this is a static template only for a test.

pawel-kow (approved):
lgtm

PR #481, opened 2023-12-19
MailerSend template v2

fosron:
We are updating our DKIM from TXT to CNAME and we have to have 2 templates for this to work.


pawel-kow (approved):
lgtm

PR #483, opened 2023-12-27
canva: Update TXT verification name in canva.com.website.json

tomnugent:


pawel-kow (dismissed):
Lgtm

tomnugent:
Hey team, I'm just waiting on a new logo and will update the description accordingly

canva.com.website.json
```diff
@@ -4,26 +4,29 @@
...
    "syncPubKeyDomain":"domainconnect.canva.com",
-   "version": 2,
-   "logoUrl": "https://static.canva.com/static/images/canva_logo_100x100@2x.png",
+   "version":3,
+   "logoUrl":"https://static.canva.com/static/images/canva_logo_100x100@2x.png",
    "description":"Enables a domain to work with Canva",
    "variableDescription":"verify is the verification value for the site",
    "records":[  
       {  
+         "groupId":"hosting",
```
tomnugent, line 13: This allows our FE to request only for hosting or verification records to be set. I don't think we'll use it yet, but we may as well add it here while we update the template.

See [here](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#template-record:~:text=in%20the%20template.-,Group,-Id)
> (optional) This parameter specifies the groups from the template to apply. If no group is specified, all groups are applied. Multiple groups may be specified in a comma delimited format.

canva.com.website.json
```diff
@@ -4,26 +4,29 @@
    "serviceId":"website",
    "serviceName":"Canva Websites",
    "syncPubKeyDomain":"domainconnect.canva.com",
-   "version": 2,
-   "logoUrl": "https://static.canva.com/static/images/canva_logo_100x100@2x.png",
+   "version":3,
+   "logoUrl":"https://static.canva.com/static/images/canva_logo_100x100@2x.png",
```
tomnugent, line 8: waiting to update this

canva.com.website.json
```diff
@@ -4,27 +4,37 @@
...
-         "data": "%verify%",
+         "host":"@",
+         "data":"%verify%",
+         "ttl":300
+      },
+      {
+         "groupId":"verificationCname",
+         "type":"CNAME",
+         "host":"canva-domain-verify",
+         "pointsTo":"%verify%",
```
tomnugent, line 37:
```suggestion
         "host":"%verify-name%",
         "data":"%verify-value%",
         "ttl":300
      },
      {
         "groupId":"verificationCname",
         "type":"CNAME",
         "host":"%verify-name%",
         "pointsTo":"%verify-value%",
```

kerolasa:
Looks good to me. Strange this is still waiting for a merge.

tomnugent:
Hey @kerolasa, apologies, we've been waiting on some changes in our own codebase to make onboarding new DomainConnect templates more seamless. Is it preferable that we close this PR without merging and then open a new PR when we're ready? Otherwise I can keep it as a Draft for now

canva.com.website.json
```diff
@@ -4,27 +4,37 @@
...
          "host":"www",
          "pointsTo":"103.169.142.0",
          "ttl":300
       },
       {
+         "groupId":"verificationTxt",
          "type":"TXT",
-         "host": "_canva-domain-verify",
-         "data": "%verify%",
+         "host":"%verify-name%",
```
tomnugent, line 29: Are we able to specify `@` for this variable?

e.g. `https://web-connect.dnsprovider.com/v2/domainTemplates/providers/canva/services/template/apply?domain=example.com&verify-name=@&verify-value=123`

pawel-kow: This way it won't work. The specification does not cover for such chaining of variables.
But you can use %fqdn% directly in the template

tomnugent: Ah ok. Just to be sure I know what you mean, is it true that the query string can chain variables, like so?
`?domain=example.com&host=foo&var1=abc&var2=123`

But you're saying that we can't use `@` as a value here. Is that correct?

If that's the case, we can hardcode `"host":"@"` into our template here 👍 But for the CNAME, we'd still like to use `%verify-name%` (since it will always be defined, and will never be `@`).

Some examples of how we plan to use this template are below:

| Intended verification record | Query String |
---
| CNAME `_verify.example.com` -> `123.canva.com` | `domain=example.com&host=foo&verify-name=_verify&verify-value=123.verify.canva.com&groupId=hosting,verificationCname` |
| TXT `@` -> `verify=123` | `domain=example.com&host=foo&verify-value=verify%3D123&groupId=hosting,verificationTxt` ** |

**I'm assuming we encode `=` to `%3D` if passing it in as a variable

pawel-kow: Yes, this will work. On a side note, putting TXT record on apex has some downsides so you may want to reconsider:
https://www.ietf.org/archive/id/draft-ietf-dnsop-domain-verification-techniques-01.html#name-txt-based

BTW: you can test all of that using an online editor https://pdnsadmin.revproxy.short-lived.de/dc/free/templateedit
Just a remark - this implementation actually allows for @ in a variable - but this may be an exception.

tomnugent: Thanks @pawel-kow, I forgot about the online editor and it's perfect!

Since this template version does include breaking changes, I've opted to create a new template in https://github.com/Domain-Connect/Templates/pull/541

PR #487, opened 2024-01-12
Add brevo template as service provider

puneet60:


pawel-kow (changes requested):
With DKIM defined with pure variables I recommend using signing or warnPhishing for the template.

puneet60:
> With DKIM defined with pure variables I recommend using signing or warnPhishing for the template.

We can keep DKIM key as static. We have 2 DKIM key only. so i think we can keep 2 templates

pawel-kow (approved):
lgtm

PR #488, opened 2024-01-12
Add dmarc group

davestys:
Add dmarc group so that AWeber users can more easily publish a dmarc record if one doesn't already exist.


pawel-kow (changes requested):
Pls bump the version.

pawel-kow (approved):
lgtm

davestys:
Thanks @pawel-kow !

PR #495, opened 2024-01-23
update website dns and add email dns

limin12891:
1. providerId and serviceId not change.

2. use SPFM instead of SPF TXT Record.

### DIFF
--- a/chime.me.website.json
+++ b/chime.me.website.json
@@ -1,30 +1,84 @@
 {
-   "providerId":"chime.me",
-   "providerName":"Chime",
-   "serviceId":"website",
-   "serviceName":"Chime Real Estate Site",
-   "version": 2,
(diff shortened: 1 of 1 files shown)


chime.me.website.json
```diff
@@ -1,30 +1,85 @@
...
+      "ttl": 3600,
+      "priority": 10
+    },
+    {
+      "type": "TXT",
+      "groupId": "email",
+      "host": "_dmarc",
+      "ttl": 3600,
+      "data": "v=DMARC1; p=none",
+      "txtConflictMatchingMode": "ALL"
```
pawel-kow, line 68: As the values are case sensitive, please use `All` as per spec, not "ALL"

pawel-kow (changes requested):

chime.me.website.json
limin12891: ok, I will update it.

PR #497, opened 2024-01-31
Update Mango Mail Image

MrExcitingDigits:
Revised the logoUrl to a more appropriate image (one which includes the name instead of just the logo).


pawel-kow (approved):
lgtm

PR #501, opened 2024-02-08
Create sender.net.domain-authentication.json

edgarsendernet:


pawel-kow (changes requested):
Even though this template does not contain any variables, changing DMARC policy to p=none may lower the security level of a domain and pose a potential phishing target.
I would suggest signing of the template or adding a phishing warning flag.

edgarsendernet:
Made changes: published a public key for signing the template and created a variable for DMARC policy.

Our algorithm will work this way: before redirecting user to the DNS provider, we check for the presence of DMARC policy on the target domain. If there is one, we leave it unchanged by passing just the spf_dkim group. We created 2 groups (spf_dkim and dmarc) for this reason.

The reason for initially going with p=none is that we have noticed quite a few users who forget to authenticate other mail servers or tools that they are using, and then run into issues with mail delivery from those if DMARC is set to p=quarantine. After a transition period, we will start suggesting p=quarantine by default.

pawel-kow:
@edgarsendernet don't get me wrong. I don't asses whether `p=none` is right or wrong, just that there is a potential of misuse so either `syncPubKeyDomain` or `warnPhishing` should be set.
Now that you added `syncPubKeyDomain` all is ok on my side. You may also go with static `p=none` or keep the variable as it is now. Just let me know if the template is final for merge now.

edgarsendernet:
The variable will let us be more flexible when rolling out p=quarantine.

The template is final, can be merged, thanks.

pawel-kow (approved):
lgtm

PR #504, opened 2024-02-28
Remove myPaymail Domain Connect template

polmyeste:
I would like to remove the myPaymail Domain Connect template as the site is no longer operational.


pawel-kow (approved):
lgtm 😢

PR #507, opened 2024-03-05
Add sharedProviderName parameter to Microsoft.O365 template

alinconstantin:
Add sharedProviderName parameter to Microsoft.O365 template.
This is recommended to be set for backward compatibility with DNS Providers not at V2.2 of the spec. When this was introduced 4 years ago, only the Microsoft.DPOO template was updated...


kerolasa:
Looks good to me.

wdbaker-godaddy (approved):
approved

PR #508, opened 2024-03-06
Update zoho.com.zmail_hosting.json to support multiple DKIM

dhanraj-ramasamy:
More than one DKIM records have to be updated in DNS at same time.


zoho.com.zmail_hosting.json
```diff
@@ -63,6 +63,20 @@
         "data": "%dkim_data%",
         "ttl": 600
     },
+    {
+        "groupId": "zoho-dkim1",
+        "type": "TXT",
+        "host": "%dkim_host%._domainkey",
```
pawel-kow, line 69: I would expect different variables for host and value in each of DKIM record, not `%dkim_host%` and `%dkim_data%` being the same in all three.

pawel-kow (changes requested):

zoho.com.zmail_hosting.json
dhanraj-ramasamy: I hope alpha numeric value is supported in the host and value fields. So I gave different variable name for host and value using a number. Please check

pawel-kow (approved):
yes, that's ok

PR #511, opened 2024-03-22
add getfoundation.com website template

larebnawabgetfoundation:
Add getfoundation.com website template.
Description - added a variable called "sub" to the template so that pointTo redirect to {sub}.getfoundation.store.

### DIFF
--- a/getfoundation.com.website.json
+++ b/getfoundation.com.website.json
@@ -1,0 +1,25 @@
+{
+  "providerId": "getfoundation.com",
+  "providerName": "Foundation Domain Configuration",
+  "serviceId": "foundation-template",
+  "serviceName": "Domain configuration",
+  "version": 1,
+  "logoUrl":"https://getfoundation.com/wp-content/uploads/2023/07/Foundation_logo.png",
+  "description": "Enables a domain to work with GoDaddy Whitelabel",
+  "variableDescription": "v1 is the IP address of the server.",
+  "syncPubKeyDomain": "domain-attach.api.godaddy.com",
+  "records": [
+    {
+      "pointsTo": "%v1%",
+      "type": "A",
+      "host": "@",
+      "ttl": 3600
+    },
+    {
+      "pointsTo": "@",
+      "type": "CNAME",
+      "host": "www",
+      "ttl": 3600
+    }
+  ]
+}
--- a/getfoundation.com.website.json
+++ b/getfoundation.com.website.json
@@ -1,24 +1,19 @@
 {
   "providerId": "getfoundation.com",
-  "providerName": "Foundation Domain Configuration",
-  "serviceId": "foundation-template",
-  "serviceName": "Domain configuration",
+  "providerName": "Foundation Commerce Inc",
+  "serviceId": "foundation-wholesale",
+  "serviceName": "Foundation Domain Configuration",
   "version": 1,
   "logoUrl":"https://getfoundation.com/wp-content/uploads/2023/07/Foundation_logo.png",
-  "description": "Enables a domain to work with GoDaddy Whitelabel",
-  "variableDescription": "v1 is the IP address of the server.",
-  "syncPubKeyDomain": "domain-attach.api.godaddy.com",
+  "description": "Enables the wholesale eCommerce site to work with Foundation",
+  "variableDescription":"sub is the subdomain for the domain.",
+  "syncPubKeyDomain": "getfoundation.com",
+  "syncRedirectDomain":"app.getfoundation.com",
   "records": [
     {
-      "pointsTo": "%v1%",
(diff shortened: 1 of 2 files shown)


kerolasa:
Changes looks good to me.

getfoundation.com.website.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": 1,
+  "logoUrl":"https://getfoundation.com/wp-content/uploads/2023/07/Foundation_logo.png",
+  "description": "Enables the wholesale eCommerce site to work with Foundation",
+  "variableDescription":"sub is the subdomain for the domain.",
+  "syncPubKeyDomain": "getfoundation.com",
+  "syncRedirectDomain":"app.getfoundation.com",
+  "records": [
+    {
+      "type": "CNAME",
+      "host": "%sub%",
```
pawel-kow, line 15: Even though this template would work fine for single instance, it won't work fine for multiple ones as the DNS providers would drop the other template if the `host` variable is not used for subdomains.
This looks to me as a typical one subdomain CNAME template. In this case the `%host%` built-in variable shall be used.
In this case `host` property would be `@`, `pointsTo` would be `%host%.getfoundation.store` and `hostRequired` would need to be set to `true` for this template. 
Other concern is whether you'd want for `foo.example.com` the `pointsTo` to be `foo.getfoundation.store` or rather `foo.example.com.getfoundation.store` where `%fqdn%` variable would be more appropriate.

I recommend using test editor to check how the template works.
https://pdnsadmin.revproxy.short-lived.de/dc/free/templateedit

pawel-kow (changes requested):

getfoundation.com.website.json
```diff
@@ -0,0 +1,20 @@
+{
+  "providerId": "getfoundation.com",
+  "providerName": "Foundation Commerce Inc",
+  "serviceId": "foundation-wholesale",
```
pawel-kow, line 4: AssertionError: File name getfoundation.com.website.json not matching the pattern. Expected getfoundation.com.foundation-wholesale.json
E

getfoundation.com.website.json
larebnawabgetfoundation: Thanks @pawel-kow for your detailed feedback, what exactly we need is `foo.example.com` should point to `(pointsTo)` `example.getfoundation.store`

larebnawabgetfoundation: Thanks @pawel-kow for your detailed feedback, what exactly we need is `foo.example.com` or `example.com` should point to (pointsTo) `example.getfoundation.store`

larebnawabgetfoundation: Thanks @pawel-kow for your detailed feedback, what exactly we need is the user-desired URL `foo.example.com` or 
 `example.com` should point to (pointsTo) `{subdomain}.getfoundation.store`, where this `{subdomain}` can be passed as an additional query parameter.

pawel-kow: pls explain how you expect {subdomain} to resolve? Is it just a hostname or FQDN?

I also recommend 6.6 - 6.9 of the spec with details how templates work with subdomains and `host` parameter.

BTW: this template cannot be used with just APEX domains like just example.com, as CNAME is not allowed on APEX (DNS RFC).

larebnawabgetfoundation: Hi @pawel-kow, I'll try to explain the requirements here. Please consider this example

Assume that there is a seller 'lareb' onboard on https://getfoundation.store, after onboarding he can access the portal on `https://lareb.getfoundation.store`

Since as a seller (`lareb`), I own the domain called lareb.com and now instead of `https://lareb.getfoundation.store`, I want `market.lareb.com` or `larebmarket.com` or any domain which I own? In other words `market.lareb.com` or `larebmarket.com` (anyone) should be pointed to `https://lareb.getfoundation.store`.

That's why I kept these settings for record
```
{
   "type": "CNAME",
   "host": "@",
   "pointsTo": "%sub%.getfoundation.store",
   "ttl": 3600
 }  
```

Screenshot from 2024-03-27 02-13-28
Screenshot from 2024-03-27 02-13-53

This will give you more clarity, please let me know if I am still missing anything.

getfoundation.com.website.json
larebnawabgetfoundation: Name changed.

getfoundation.com.website.json
pawel-kow: ok. this is fine. So the usage of `%sub%` variable is clear now. Thanks.
In your examples you've always used a host setting and this is correct. The template now specifies `hostRequired` to cover for that so all is good now.

pawel-kow (approved):
lgtm

PR #523, opened 2024-04-19
Add DMARC record and introduce groups in the records

puneet60:


brevo.com.domain-authentication.json
```diff
@@ -3,24 +3,41 @@
...
-        "data": "%dkim_public_key%",
-        "ttl": 3600
+      "groupId": "dkim_raw",
+      "type": "TXT",
+      "host": "%dkim_selector%._domainkey",
+      "data": "%dkim_public_key%",
+      "ttl": 3600
+    },
+    {
+      "groupId": "dmarc",
```
pawel-kow, line 28: As DMARC tends to rise a need to be edited afterwards manually there is quite useful to define this record as non essential. Just for consideration.

brevo.com.domain-authentication.json
```diff
@@ -3,24 +3,41 @@
...
   "records": [
     {
-        "type": "TXT",
-        "host": "@",
-        "data": "brevo-code:%hash%",
-        "ttl": 3600
+      "groupId": "code",
+      "type": "TXT",
+      "host": "@",
+      "data": "%hash%",
```
pawel-kow, line 17: If the semantics of a variable changes I suggest to use a new name. This makes the rollover of the template easier when some providers still have the old version and maybe do not report the version over api. Pushing both variables new and old in the call does not pose a problem.

pawel-kow (changes requested):

brevo.com.domain-authentication.json
puneet60: Done

brevo.com.domain-authentication.json
puneet60: Thanks for suggestion. Done

pawel-kow (approved):
lgtm

PR #524, opened 2024-04-30
fix: mx record

b805rohit:


pawel-kow (commented):
The change itself looks ok.

If you make an update please consider if it is not worth a while to add other `essential` settings to some of the records.
I saw it happening with other similar mail templates, that especially DMARC as shared record tends to be manually updated afterwards causing all records of the template dropped for some providers which take care of  template consistency.

pawel-kow (approved):
Lgtm

PR #530, opened 2024-05-08
Make host dynamic for DNS records to support records on subdomain

puneet60:


brevo.com.domain-authentication.json
```diff
@@ -3,31 +3,31 @@
...
-      "host": "%dkim_selector%._domainkey",
+      "host": "%dkim_host%",
       "data": "%dkim_public_key%",
       "ttl": 3600
     },
     {
       "groupId": "dmarc",
       "type": "TXT",
-      "host": "_dmarc",
+      "host": "%dmarc_host%",
```
pawel-kow, line 30: Just so I get the changes right... why would _dmarc ever be on a different host?

puneet60: We have a use case where we would add dmarc to subdomain.
For example: `_dmarc.abcd` 
In current template we would not able to add dmarc record on subdomain

pawel-kow: Unterstood. Would in this case all the other records land on the same subdomain?

puneet60: Yes all record will be on same subdomain, just difference of host.
For example dkim can be `mail._domainkey.abcd` and dmarc will be `_dmarc.abcd`

pawel-kow: I guess then this template change is not needed at all? Just use `host` parameter when applying the template.

Test yourself here: https://pdnsadmin.revproxy.short-lived.de/dc/free/templateedit

My test:
image

puneet60: Thanks a lot @pawel-kow 
That's exactly I wanted it to work. Now No further need to change template.
We can close this PR

PR #536, opened 2024-05-23
add new groupId and update A record

limin12891:
1. add new groupId "hosting";
2. update A record for groupId "website".


pawel-kow (approved):
Lgtm

PR #537, opened 2024-05-27
update lofty.com.website.json

limin12891:
1、update groupId="website" from Type "A" to Type "CNAME";

2、update groupId="email" from SPFM records to a normal TXT, because godaddy provider do not support non-apex SPFM records.

### DIFF
--- a/lofty.com.website.json
+++ b/lofty.com.website.json
@@ -3,7 +3,7 @@
   "providerName": "Lofty",
   "serviceId": "website",
   "serviceName": "Lofty Real Estate Website",
-  "version": 3,
+  "version": 4,
   "logoUrl": "https://static.chimeroi.com/servicetool-temp/2023910/17/crm/lofty-Logo.svg",
   "description": "Lofty customers can use their own domain name or subdomain to access their website built on lofty",
   "syncRedirectDomain": "lofty.com",
@@ -23,18 +23,11 @@
       "ttl": 3600
     },
     {
-      "type": "A",
+      "type":"CNAME",
       "groupId": "website",
-      "host": "%sub%",
-      "pointsTo": "52.52.24.52",
-      "ttl": 3600
-    },
-    {
-      "type": "A",
-      "groupId": "website",
-      "host": "%sub%",
-      "pointsTo": "52.9.101.47",
-      "ttl": 3600
+      "host":"%sub%",
+      "pointsTo": "site-dns.chime.me",
+      "ttl":3600
     },
     {
       "type": "CNAME",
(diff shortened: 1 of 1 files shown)


lofty.com.website.json
```diff
@@ -82,16 +75,17 @@
...
     {
-      "type": "SPFM",
+      "type": "TXT",
       "groupId": "email",
-      "host": "@",
+      "host": "mail",
       "ttl": 3600,
-      "spfRules": "include:amazonses.com"
+      "data": "v=spf1 include:amazonses.com ~all",
+      "txtConflictMatchingMode": "All"
```
pawel-kow, line 83: This would prevent any TXT records on this host. For SPF actually prefix would be the right approach.

lofty.com.website.json
```diff
@@ -82,16 +75,17 @@
...
     },
     {
-      "type": "SPFM",
+      "type": "TXT",
       "groupId": "email",
-      "host": "@",
+      "host": "mail",
       "ttl": 3600,
-      "spfRules": "include:amazonses.com"
+      "data": "v=spf1 include:amazonses.com ~all",
```
pawel-kow, line 82: If it's only GoDaddy I don't like in particular then affecting all other providers. Even if this is a subdomain the impact may be smaller but nevertheless not non existing. 

Better approaches:
- special group for GoDaddy variant
- let GoDaddy transform the template during onboarding
- special separate template

lofty.com.website.json
```diff
@@ -82,16 +75,17 @@
       "txtConflictMatchingMode": "All"
     },
     {
-      "type": "SPFM",
+      "type": "TXT",
```
pawel-kow, line 78: Like essential setting is missing.  See also 6.10.4 in the spec.

pawel-kow (changes requested):

limin12891:
I update lofty.com.website.json with reference to 6.10.4 in the spec, please review it again.

lofty.com.website.json
```diff
@@ -82,18 +75,20 @@
...
       "groupId": "email",
-      "host": "mail",
+      "host": "@",
       "ttl": 3600,
-      "spfRules": "include:amazonses.com"
+      "data": "v=spf1 %spfRules% ~all",
```
pawel-kow, line 90: Replacing SPFM on APEX is actually a bad idea. Do it only if you plan to make a full merge of SPF on your side and even if you do that there are field reports of very bad side effects.
Normally you'd be better off sticking to SPFM. This approach is just a workaround (and actually in the spec how DNS provider may emulate SPFM).

limin12891: ok, I will use SPFM record on APEX.

pawel-kow (approved):
ok

PR #539, opened 2024-06-06
MailerSend template - Support subdomain for DKIM TXT record

JoaoFSCruz:
Currently we cannot add the subdomain to the DKIM TXT record host because we hardcoded the `._domainkey` suffix. We need to remove it and handle that part entirely on our side.


paulokow:
is the template deployed (widely) so that you need to roll the version?
if yes you'd rather have a new variable and not change the semantics of an existing one.

JoaoFSCruz:
Hey @paulokow! Thanks for replying. Currently we only support Cloudflare and GoDaddy. Maybe we do not require to roll the version?

pawel-kow:
> Hey @paulokow! Thanks for replying. Currently we only support Cloudflare and GoDaddy. Maybe we do not require to roll the version?

Depends on you. You either make then new/old semantics based on the template version from the API, or make 2 variables and push both of them until the transition is done.

Typically you don't know exactly when a provider will deploy the new version and it's rather difficult to synchronise 2 or more providers.

Choice is yours. Let me know.

JoaoFSCruz:
Thanks for your input but we'll postpone this change for sometime. 🙇

PR #540, opened 2024-06-12
Added Upronix Template JSON

slxca:


upronix.com.domain-setup.json
```diff
@@ -0,0 +1,17 @@
...
+  "providerName": "Upronix",
+  "serviceId": "domain-setup",
+  "version": 1,
+  "serviceName": "Domain Setup",
+  "logoUrl": "https://upronix.com/domain-connect/logo.png",
+  "description": "Setup an Domain to use Custom Domains",
+  "records": [
+    {
+      "type": "CNAME",
+      "host": "%domain%",
```
pawel-kow, line 12: `%domain%` is a reserved variable, so in the end this would mean creating a subdomain with the same hostname as domain. Like `example.com.example.com`.
I don't think this was the intention.
I guess this template is for subdomains, then right thing to do would be host=@ and `hostRequired` set to `true`. Then the right host would be set with `host` parameter.

Anyway, please test your template here https://pdnsadmin.revproxy.short-lived.de/dc/free/templateedit and provide an appropriate update.

pawel-kow (changes requested):

pawel-kow (approved):
lgtm

PR #541, opened 2024-06-12
Add Canva website-v2 template

tomnugent:
## This PR

This PR adds a `canva.com.website-v2.json` template, which builds on [canva.com.website.json](https://github.com/Domain-Connect/Templates/blob/1e9773bd2ba57d93e3109876f4805dbef32067f1/canva.com.website.json) but introduces breaking changes. A new template is being introduced based on advice in [the Spec here](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#51-template-versioning).

**More context / discussion:** https://github.com/Domain-Connect/Templates/pull/483

This template allows us to:
- set TXT ownership record on any name, including the root (`@`)
- specify `groupId`s to allow setting hosting records (A records), TXT, or CNAME verification records separately
- use a different logo

## Examples

| Intended verification record | Query String |
| --- | --- |
| CNAME `_verify.example.com` -> `123.canva.com` | `domain=example.com&verify-cname-name=_verify&verify-cname-value=123.canva.com&groupId=hostingA,verificationCname` |
| TXT `@` -> `verify=123` | `domain=example.com&verify-txt-value=verify%3D123&groupId=hostingA,verificationTxt`** |

**Using `%3D` as the `=` symbol.

## Test

Using https://pdnsadmin.revproxy.short-lived.de/dc/free/templateedit#tbl_records_and_errors

Screenshot 2024-06-13 at 11 40 12 AM


tomnugent:
This is now ready for review, thanks in advance!

canva.com.website-v2.json
```diff
@@ -0,0 +1,55 @@
...
+         "groupId":"hostingAAAA",
+         "type":"AAAA",
+         "host":"www",
+         "pointsTo":"%ipv6%",
+         "ttl":3600
+      },
+      {
+         "groupId":"verificationTxt",
+         "type":"TXT",
+         "host":"%verify-txt-name%",
```
pawel-kow, line 43: What's the point of this variable if you don't intend to use it or leave it empty?

tomnugent: Hey @pawel-kow, we want to have the option of using a different name if the need arises. This way we won't need to make a new template version if we ever change our minds on the verification record format.

pawel-kow: ok, fair enough.

pawel-kow (approved):

PR #543, opened 2024-06-14
Update upronix.com.domain-setup.json

slxca:


upronix.com.domain-setup.json
```diff
@@ -2,17 +2,19 @@
...
   "logoUrl": "https://upronix.com/domain-connect/logo.png",
-  "description": "Setup an Domain to use Custom Domains",
-  "hostRequired": true,
+  "description": "Setup a Domain to use Custom Domains",
   "records": [
     {
+      "groupId": "setup",
       "type": "CNAME",
-      "host": "@",
+      "host": "%setuphost%",
```
pawel-kow, line 15: This change does not make any sense. This is the only variable and used the same way as host parameter.
It looks like trying to overcome hostRequired limitation on CNAME on APEX. This won't work anyway as DNS forbids it, so original template was correct in this sense.
Please explain what you try to gain with this change.

pawel-kow (changes requested):

upronix.com.domain-setup.json
slxca: I'm trying to create a subdomain as CNAME. Example: sub.example.com. However, Cloudflare does not allow the "hostRequired" field.

pawel-kow: @slxca the original template was actually how it was supposed to be.

This is the Cloudflare linter output. The only issue was `ERR` on missing `syncPubKeyDomain`. `hostRequired` may be not supported but I would assume that it's just ignored and not forbidden. There are many templates with CNAME on @, that I don't believe it would be true. @kerolasa may you please confirm/deny :)
```
$ dc-template-linter -cloudflare -pretty upronix.com.domain-setup.json
2024-06-14T20:25:40+02:00 ERR code=DCTL5001 dctl_note="syncPubKeyDomain is required" template=upronix.com.domain-setup.json
2024-06-14T20:25:40+02:00 INF code=DCTL5006 dctl_note="hostRequired is not supported" template=upronix.com.domain-setup.json
2024-06-14T20:25:40+02:00 INF code=DCTL5007 dctl_note="domains must use Cloudflares CNAME flattening setting" groupid= record=0 template=upronix.com.domain-setup.json type=CNAME
```

pawel-kow (approved):
Lgtm

PR #550, opened 2024-08-05
Add KeyAuth.cc

wnelson03:
We would like to allow our customers to use DomainConnect to seamlessly setup their custom domains for our service's API.

https://keyauthdocs.apidog.io/doc-446443

Thank you for your time, I appreciate it.

image


pawel-kow (approved):
lgtm

PR #554, opened 2024-08-30
[SmugMug] fix redirect target

bwg:
target needs to be a url, not just a domain.  SmugMug provides certs, so hard-coding https is fine.

https://github.com/Domain-Connect/Templates/pull/552#discussion_r1735312593


pawel-kow (approved):
lgtm

PR #555, opened 2024-09-09
update syncRedirectDomain to be FQDN

bwg:
This needs to be a fully qualified domain, not a root domain, according to GoDaddy.  The spec is ambiguous here.


pawel-kow (approved):
lgtm

PR #556, opened 2024-09-18
Add domchekr.com.domaincheck.json template

emmanuelgautier:
Add a template verification for domchekr.com. This template will be used with several DNS Providers including Cloudflare for domains verification.


pawel-kow:
@emmanuelgautier is this template going to be applied more than once? If yes think of `multiInstance` setting.
If not and it would be required to remove the previous one before adding a new one take a look whether `txtConflictMatchingMode=Prefix` wouldn't be of a value to you.

emmanuelgautier:
@pawel-kow You are right, the template may be applied more than once for the same host. `multiInstance` is now set to `true`. Thanks for your review.

pawel-kow (approved):
lgtm

PR #560, opened 2024-10-07
Add TinkerMail by TinkerHost Service

greenreader9:
Adds Domain Connect support to upcoming TinkerMail service


greenreader9:
Changed to DRAFT as the template may be updated within the next 48 hours

kerolasa:
Something is wrong with linter check. It did not notice file name is missing `.json`

```
Templates $ dc-template-linter -inplace tinkerhost.net.tinkermail 
2024-10-07T15:04:47+01:00 ERR code=DCTL1003 dctl_note="file name does not use required pattern" expected=tinkerhost.net.tinkermail.json template=tinkerhost.net.tinkermail
```

greenreader9:
Forgetting the JSON extension is a bit embarrassing, let me fix that right now

greenreader9:
Fixed @kerolasa

pawel-kow:
> Forgetting the JSON extension is a bit embarrassing, let me fix that right now

I think there is an issue with the github action as it selects only json files so any other file won't hit the linter at all
```
      - name: Get all changed json files
        id: changed-json-files
        uses: tj-actions/changed-files@v42
        with:
          files: |
             **.json
```

I think here we'd need to either let all files to run through linter, even if these are not related (README) or to have some smarter solution with exclusion lists...
Alternatively we may leave it for manual review if any file is not json. This should be quite easy to notice...

kerolasa:
@greenreader9 the template looks good to me.

Meanwhile, I think the following change to github actions might work. Should we try @pawel-kow? 

```
diff --git a/.github/workflows/dc-template-lint.yml b/.github/workflows/dc-template-lint.yml
index 5d85d7c..3fb3d65 100644
--- a/.github/workflows/dc-template-lint.yml
+++ b/.github/workflows/dc-template-lint.yml
@@ -35,7 +35,9 @@ jobs:
         uses: tj-actions/changed-files@v42
         with:
           files: |
-             **.json
+             '**',
+             '!README.md',
+             '!template.schema'
 
       - name: Print all changed json files
         if: steps.changed-json-files.outputs.any_changed == 'true'
```

pawel-kow (approved):
lgtm

PR #565, opened 2024-10-26
Add KeyAuth.com

wnelson03:
We would like to allow our customers to use DomainConnect to seamlessly setup their custom domains for our service's self user-management portal.

Example: https://panel02.nelsoncybersecurity.com

I confirm it's intentional that `syncPubKeyDomain` and `logoUrl` don't match exactly to `providerId`. Cloudflare's custom hostnames for SaaS product doesn't appear to allow multiple different use-cases, it only allows 1 use-case per domain which has already been used.

Thank you for your time, I appreciate it.

image


pawel-kow (approved):
Lgtm

PR #567, opened 2024-11-03
Update tinkerhost.net.tinkermail.json to version 2

greenreader9:
Adds support for TXT record _dmarc. Does not add or change any variables.

### DIFF
--- a/tinkerhost.net.tinkermail.json
+++ b/tinkerhost.net.tinkermail.json
@@ -57,6 +57,15 @@
       "host": "_dmarc",
       "ttl": 3600,
       "data": "v=DMARC1;p=none"
+    },
+    {
+      "type": "TXT",
+      "essential": "onApply",
+      "host": "_dmarc",
+      "ttl": 3600,
+      "data": "v=DMARC1;p=none",
+      "txtConflictMatchingMode": "Prefix",
+      "txtConflictMatchingPrefix": "v=DMARC1;"
     }
   ]
 }
--- a/tinkerhost.net.tinkermail.json
+++ b/tinkerhost.net.tinkermail.json
@@ -54,12 +54,6 @@
     },
     {
       "type": "TXT",
-      "host": "_dmarc",
-      "ttl": 3600,
-      "data": "v=DMARC1;p=none"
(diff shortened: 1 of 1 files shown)


greenreader9:
Is there a SPFM-like record type that can be used for DMARC records as they are also restricted to one per hostname?

pawel-kow:
@greenreader9 setting `txtConflictMatchingMode` to `All` is intended for this purpose. It's not (yet) widely supported by DNS providers.
Other (intermediate) option would be to put CNAME there which would always wipe anything else.

greenreader9:
@pawel-kow, if txtConflictMatchingMode and txtConflictMatchingPrefix were added to the record and the DNS provider does not support it, I would assume it would just get overridden by the provider?

I don't want to override a users DMARC record if they already have one in place (especially since my default is "none"), and placing a CNAME record to a DMARC record on a domain where a DMARC record is already in place probably constitutes multiple DMARC records for the domain.

pawel-kow:
> @pawel-kow, if txtConflictMatchingMode and txtConflictMatchingPrefix were added to the record and the DNS provider does not support it, I would assume it would just get overridden by the provider?

No, unfortunately not. They will just add a second record.
CNAME would guarantee a replacement as CNAME cannot coexist with any other record, so the provider will always replace.
 
> I don't want to override a users DMARC record if they already have one in place (especially since my default is "none"), and placing a CNAME record to a DMARC record on a domain where a DMARC record is already in place probably constitutes multiple DMARC records for the domain.

The solution to that would be to define a group for DMARC and apply this group only if DMARC is not yet present. This would also solve the problem of the duplicate if the intention is to only add DMARC if there is none present.
I would also suggest adding essential=OnApply so that DMARC can be updated/replaced without dropping the whole template.

greenreader9:
@pawel-kow , updated the record to include both 'essential' as well as 'txtConflictMatchingMode' for supported providers

kerolasa:
Would it be possible to squash commits 5feaa01 and 4288bb3 together.

pawel-kow (approved):
lgtm!

PR #576, opened 2025-01-31
Create vercel.com template

RhysSullivan:
# Description

Hey, we're setting up domain connect to make it easier for our users to add domains

When a user is configuring a domain on Vercel, there's 3 fields they need to set:

1. TXT Verification - The user may need to do this to prove ownership of their domain, optional
2. CNAME Record - Only used if the user is adding a subdomain
3. A Record - Only used if the user is adding an apex domain

Here's an example of the types of states the domain config can be in:
image

I'm also working on getting us setup as a DNS Provider with Domain Connect support as well

Thanks for taking a look at the PR, let me know any fixes that need to be made

## Type of change

- [X] New template

# How Has This Been Tested?

- [X] Schema validated using JSON Schema [template.schema](./template.schema)
- [X] Template is checked using [template linter](https://github.com/Domain-Connect/dc-template-linter)
- [X] Template file name follows the pattern `<providerId>.<serviceId>.json`

# Example variable values

```
verification: cloudflare.com,99d313170ac2cfc82c10
subdomain: www
```

### DIFF
--- a/vercel.com.website.json
+++ b/vercel.com.website.json
@@ -9,18 +9,21 @@
   "records": [
     {
       "type": "TXT",
+      "groupId": "verification",
       "host": "_vercel",
       "data": "vc-domain-verify=%verification%",
       "ttl": 3600
     },
     {
       "type":"CNAME",
+      "groupId": "subdomain",
       "host": "%subdomain%",
       "pointsTo": "cname.vercel-dns.com",
       "ttl":3600
     },
     {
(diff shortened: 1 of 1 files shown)


kerolasa (approved):
Three commits could be squashed to single one, but that's a minor nit. The template data itself looks good to me.

pawel-kow (approved):
lgtm!

PR #586, opened 2025-02-24
Create Platelink templates

shubhusion:
# Description

Templates for adding a custom domain/subdomain to a Platelink site

## Type of change

- [X] New template

# How Has This Been Tested?

- [X] Schema validated using JSON Schema [template.schema](./template.schema)
- [X] Template is checked using [template linter](https://github.com/Domain-Connect/dc-template-linter)
- [X] Template file name follows the pattern `<providerId>.<serviceId>.json`

### DIFF
--- a/platelink.dev.domain.json
+++ b/platelink.dev.domain.json
@@ -1,48 +1,31 @@
 {
-    "providerId": "platelink.dev",
-    "providerName": "Platelink",
-    "serviceId": "domain",
-    "serviceName": "Platelink Custom Domain [DEV]",
-    "version": 1,
-    "logoUrl": "https://stage-hub.platelink.ai/logo.svg",
-    "description": "Enables domains and subdomains to work with Platelink",
-    "syncPubKeyDomain": "signingkey.platelink.ai",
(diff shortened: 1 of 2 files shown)


pawel-kow:
I don't think it was right suggestion @kerolasa to join the template.
Now the new template subdomain part cannot use host parameter to create subdomains, because CNAME cannot use be on APEX without hostRequired set, which would not work for the main domain. Two templates was the right approach AFAIK

shubhusion:
> I don't think it was right suggestion @kerolasa to join the template. Now the new template subdomain part cannot use host parameter to create subdomains, because CNAME cannot be on @ without hostRequired set, which would not work for the main domain. Two templates was the right approach AFAIK

@pawel-kow so should I proceed with my previous two templates ?

pawel-kow:
> > I don't think it was right suggestion @kerolasa to join the template. Now the new template subdomain part cannot use host parameter to create subdomains, because CNAME cannot be on @ without hostRequired set, which would not work for the main domain. Two templates was the right approach AFAIK
> 
> @pawel-kow , should I proceed with my previous two templates then? Let me know what you think is the best way to move forward.

Yes please.

shubhusion:
I have implemented the changes and reverted to the two-template approach as discussed. Let me know if you need any further modifications.

pawel-kow (approved):
lgtm

PR #609, opened 2025-05-13
Add more dns records like spf, a, mx and cname in template

parveen-sib:
Required for domain association with IP

### DIFF
--- a/brevo.com.domain-authentication.json
+++ b/brevo.com.domain-authentication.json
@@ -8,7 +8,7 @@
   "syncRedirectDomain": "app.brevo.com",
   "logoUrl": "https://static.brevo.com/meetings/assets/brevo_logo.png",
   "description": "Places TXT record for domain verification and DKIM records to authenticate email sent by Brevo on behalf of the user",
-  "variableDescription": "The variable %verfication_code% represents a combination of prefix string and a MD5 hash. %dkim_selector% and %dkim_selector2% is the DKIM selector for placing the DKIM public key. This needs to be a variable because the selector is dynamically generated. %dkim_public_key% is the DKIM public key. %dmarc_value% is dmarc value. %dkimcname% and %dkimcname2% is cname value where DKIM will point",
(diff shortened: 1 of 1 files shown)


brevo.com.domain-authentication.json
```diff
@@ -46,6 +46,45 @@
...
+      "type": "A",
+      "host": "@",
+      "pointsTo": "%ip%",
+      "ttl": 3600
+    },
+    {
+      "groupId": "spf",
+      "type": "TXT",
+      "host": "@",
+      "data": "%spf_value%",
```
pawel-kow, line 61: Please consider using SPFM instead or justify why not.

brevo.com.domain-authentication.json
```diff
@@ -46,6 +46,45 @@
...
+      "ttl": 3600,
+      "txtConflictMatchingMode": "Prefix",
+      "txtConflictMatchingPrefix": "v=spf1",
+      "essential": "OnApply"
+    },
+    {
+      "groupId": "mx",
+      "type": "MX",
+      "host": "@",
+      "pointsTo": "%domain%",
```
pawel-kow, line 71: It likely has to be an absolute reference, so with a dot suffix.
Also please test the template with host parameter set to see if you really want %domain% or rather `@`/`%fqdn%`

pawel-kow (changes requested):

brevo.com.domain-authentication.json
parveen-sib: @pawel-kow Yes we need to set domain name and it is already tested

pawel-kow: So trailing dot is obviously missing (see: https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc#template-record)

brevo.com.domain-authentication.json
parveen-sib: Added SPFM record

brevo.com.domain-authentication.json
parveen-sib: Update with fqdn

pawel-kow (approved):
lgtm

PR #612, opened 2025-05-14
feat: add new variable to Resend template

isabellaaquino:
# Description
We are introducing the capability of users to choose their own custom return-path address at Resend.

With that, we'll need to make the host field for these records a variable to match this new behavior (previously fixed as `send`).

## Type of change

- [ ] New template
- [x] New feature (non-breaking change which adds functionality to the template)

# How Has This Been Tested?

- [ ] Schema validated using JSON Schema [template.schema](./template.schema)
- [ ] Template is checked using [template linter](https://github.com/Domain-Connect/dc-template-linter)

# Example variable values
```
spfDomain: bounces
region: us-east-1
domainKey: RANDOM-VALUE
```

### DIFF
--- a/resend.com.mail.json
+++ b/resend.com.mail.json
@@ -3,7 +3,7 @@
 	"providerName": "Resend",
 	"serviceId": "mail",
 	"serviceName": "Mail",
-	"version": 2,
+	"version": 3,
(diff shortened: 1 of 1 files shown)


pawel-kow (approved):
Lgtm

PR #615, opened 2025-05-20
Update informaten.com.hosting_web.json

Luxaaa:
# Description

update informaten.com_hosting_web.json to use %tqdn% instead of @ in the pointsTo field because cloudflare points to the root domain if we use @.
## Type of change

- [ ] New template
- [x] Bug fix (non-breaking change which fixes an issue in the template)

# How Has This Been Tested?

- [x] Schema validated using JSON Schema [template.schema](./template.schema)
- [x] Template is checked using [template linter](https://github.com/Domain-Connect/dc-template-linter)

### DIFF
--- a/informaten.com.hosting_web.json
+++ b/informaten.com.hosting_web.json
@@ -3,7 +3,7 @@
   "providerName": "Informaten",
   "serviceId": "hosting_web",
   "serviceName": "Webhosting on Informaten.com",
-  "version": 1,
+  "version": 2,
   "syncPubKeyDomain": "informaten.com",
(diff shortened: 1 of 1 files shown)


Luxaaa:
Is there anything else I need to do here?

kerolasa:
I  brought up the `@` handling with Cloudflare folks who maintain web api component. To me it looks like Cloudflare is not doing right things with this one. I will add more info when there is something to comment.

Meanwhile, I'm ok merging the `%fqdn%` change as it should go around the issue without causing problems with other providers.

PR #635, opened 2025-08-20
docs: update template examples

kerolasa:
A while back TTL and various other number values were allowed to be variables.  This caused these numbers become strings, which is reflected now in examples.

While on it update syncPubKeyDomain and syncRedirectDomain in README to have more realistic values.


pawel-kow (changes requested):
> Version bump needed for the files:
> 
> ```
> exampleservice.domainconnect.org.template1.json
> exampleservice.domainconnect.org.template2.json
> exampleservice.domainconnect.org.test-template.json
> ```

@kerolasa pls indeed bump the versions in the example service templates.

kerolasa:
> @kerolasa pls indeed bump the versions in the example service templates.

Good point, I totally forgot examples need version updates as much as any other changes. Done in c60cb97...1e19d1d

PR #643, opened 2025-09-11
Add redirect.pizza

mbardelmeijer:
# Description

This adds 3 templates for the SaaS solution [redirect.pizza](https://redirect.pizza). redirect.pizza provides a full-domain redirect service. Redirects can be configured on the apex and/or any subdomain. This PR includes 3 templates, one for the apex (i.e. non-www to www redirect for instance), one for subdomain (i.e. test.example.com for instance), and 1 "common" case: the apex & www combined, so that the user can apply the template in 1 click for both cases.

## Type of change

- [X] New template

# How Has This Been Tested?

- [ ] Schema validated using JSON Schema [template.schema](./template.schema)
- [ ] Template is checked using [template linter](https://github.com/Domain-Connect/dc-template-linter)
- [ ] Template file name follows the pattern `<providerId>.<serviceId>.json`

# Example variable values
```
ipv4: 89.106.200.1
ipv6: 2a12:5240::1
cname: edge.redirect.pizza
```

Typically, these variables are always the same, but users may set-up a Dedicated IP in their account, which will change these values. See https://redirect.pizza/support/dns-type & https://redirect.pizza/support/dedicated-ip

Open for suggestions for the serviceId names.

### DIFF
--- a/redirect.pizza.apex-www.json
+++ b/redirect.pizza.apex-www.json
@@ -2,7 +2,7 @@
   "providerId": "redirect.pizza",
   "providerName": "redirect.pizza",
   "serviceId": "apex-www",
-  "serviceName": "redirect.pizza apex-www",
+  "serviceName": "redirect.pizza",
   "version": 1,
   "syncPubKeyDomain": "domainconnect.redirect.pizza",
   "logoUrl": "https://redirect.pizza/favicons/apple-touch-icon.png",
--- a/redirect.pizza.apex.json
+++ b/redirect.pizza.apex.json
@@ -2,7 +2,7 @@
   "providerId": "redirect.pizza",
   "providerName": "redirect.pizza",
   "serviceId": "apex",
-  "serviceName": "redirect.pizza apex",
+  "serviceName": "redirect.pizza",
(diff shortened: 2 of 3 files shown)


pawel-kow (approved):
Templates with and without www could have been one with groupIds. Technically ok though.
Let me know if you want any correction before merge.

redirect.pizza.apex-www.json
```diff
@@ -0,0 +1,32 @@
+{
+  "providerId": "redirect.pizza",
+  "providerName": "redirect.pizza",
+  "serviceId": "apex-www",
+  "serviceName": "redirect.pizza apex-www",
```
pawel-kow: Please note that serviceName may/will be shown by DNS provider to the user in consent screen.
Please consider if "redirect.pizza apex-www" is the right name.

On a side note: it is ok for all 3 template variations to have same name, if user does not need to know which template is used.

I put a template on "change required" to get your response, but it's fine if you confirm to keep it as-is.

pawel-kow (changes requested):

PR #645, opened 2025-09-16
Add Pending property page template

noelxroberts:
This PR adds a Domain Connect template for Pending's property page hosting service.
   
   The template configures:
   - CNAME records for www and @ (apex) domains
   - Points to CloudFront distribution for property page hosting
   - Supports custom domain setup for property listings
   
   Provider: pending
   Service: property-page

### DIFF
--- a/pending.property-page.json
+++ b/pending.property-page.json
@@ -1,0 +1,25 @@
+{
+  "providerId": "pending",
+  "providerName": "Pending",
+  "serviceId": "property-page",
+  "serviceName": "Property Page Hosting",
+  "version": 2,
+  "logoUrl": "https://trypending.com/logo.png",
+  "description": "Enables custom domain hosting for property pages.",
+  "variableDescription": "",
+  "records": [
+    {
+      "type": "CNAME",
+      "host": "www",
+      "pointsTo": "d1234567890.cloudfront.net",
+      "ttl": 3600
+    },
+    {
+      "type": "CNAME",
+      "host": "@",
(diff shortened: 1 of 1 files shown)


kerolasa:
Looks ok, but I do want to encourage adding `syncPubKeyDomain` to the template. Many of DNS Providers, such as Cloudflare, require digitally signed updates.

Assuming the key gets to be added please also consider rebasing and squashing three commits to be a single one on top of the most recent origin/master. Commands below should give a hint how to complete this task.

```
git fetch --all &&
git checkout noelxroberts-patch-1 &&
git rebase -i origin/master && # change two last 'pick' lines to 'squash' and fix commit message
git push yourremote -f noelxroberts-patch-1:noelxroberts-patch-1
```

pawel-kow:
@noelxroberts I would appreciate fully filled in PR template, as it makes sure some testing was done upfront and saves work for unneeded reworks of the template.
Also, I notice that provider does not follow the domain name convention.
As far as it is a convention it is not a showstopper, but offers very good clarity so it would be good to know the reason.

pawel-kow (approved):
lgtm!

pawel-kow:
> Looks better, but there are more commits to be squashed. IMHO it would be nice if adding a template could be single commit which you can now achieve by using interactive rebase again.

@kerolasa there is squash merge strategy set up for the main branch on GH, so every PR will be squashed automatically when PR is merged. See 8f8701ed627919740c35b34bf88e8fa96d58ae98 for this PR as example.

PR #799, opened 2026-03-04
Update CNAME target for mcp-use custom domains

tonxxd:
Updates the CNAME record target from `gateway.mcp-use.run` to `deploy.mcp-use.com` to reflect our current gateway infrastructure.


mcp-use.com.custom-domain.json
pawel-kow, line 6: bump version

PR #955, opened 2026-04-07
Update vercel website template to add apex CNAME record

elliotdauber:
# Description

Vercel needs to be able to set CNAMEs for apex domains on providers that support CNAME flattening so that we can determine the correct IP dynamically for the domain instead of hardcoding it during DNS setup.

## Type of change

- [ ] New template
- [x] New feature (non-breaking change which adds functionality to the template)

# How Has This Been Tested?

- [x] no variable is used as a bare full record value (e.g. `@ TXT "%foo%"`) unless necessary — prefer `@ TXT "service-foo=%foo%"`; if bare, justify in the PR description
- [x] `essential` is set to `OnApply` on records the end user may need to modify or remove without breaking the template (e.g. DMARC)

## Online Editor test results

**Editor test link(s):** 

[Test vercel.com/website example.com/elliot](https://domainconnect.paulonet.eu/dc/free/templateedit?token=H4sIAE5U1WkC%2F%2B1V72vbMBD9V4ygn2oHx0nTxFBY6cp%2BlJZSQtetBKPIl1arY2mWHDcL%2Bd93luM4bmI2ygYr5JNt3bt370l38oJomMqIaiD%2BgshEzHgIyaeQ%2BGQGCYOoxcSU2OvIFZ0iktyaGK4rSGacgUnIYKw4Eq1Xa2DrSxFWGEdqxUVM%2FK5NQlAs4VKbb3Ie03EEyqJWKKaUx5YWViaSJyvj%2BtGqys5jdp2OL2D%2B3sAws8AzEcfAdKsmPkffQMgTjKzxNcSjUPoGfqQIQSc6ScEmiBZJqIh%2FvyB6LnMnw7shgh8SkUrjmEp4dpCITzijxkFBhaFgVmoNqaZ5PeYUEouE%2BcnBVvYBorWOiN9z3aXdWFWl402m15TeTdFQ%2F%2Bzq9PJ8t4KqaEWZs0jBY62GIg%2BwGNuggfp0azsrxncveLj8Y31mZ03dZroKU6MdLW3yU8QQVMc%2FslfNhWnwTHFcYKNt8sUo4kKXCgJusnY2R%2BPRbe7ohnosLWlCpyqfzlLEfU3FqJRxX%2BoYrYT8RRHbNMiNs2yKNfCtEF4NUpdZ8OOSeVkNrRPGqnTGjYm1gt%2Fg8ej4QywSCBQ%2BqU4TKId5mkaaBzSj1VLIAiplNMeTVhhF2peDXlQr56m1PuamsTIbUvWSTYKQ5QfHTU%2F2vXav3x84YR9cp9vudpwBPcI31j4%2BctvHEHp046LdvoJ3X7UvWxCUglhzGuXTFWV0rshy6y55jTHvP3dW3gIrb0XalrWdjfMGjb1BRyMbb9L9iO1HbD9i%2F2zE8l8tnUEY0BzsuV7PcbuOezxsD3zX9Ttuqz8Y9Lq9Q%2Fxw3dwDtl3hcoF%2Fz83%2FJumd3fU%2FQvp9cHjmPd5eTDowOf387cNDp%2FP1aagEj6Q8F5nLnrLLE7L8BU53U%2B5JDAAA)

### DIFF
--- a/vercel.com.website.json
+++ b/vercel.com.website.json
@@ -7,6 +7,7 @@
   "description": "Enables a domain to work with Vercel",
   "syncPubKeyDomain": "domainconnect.vercel.com",
   "syncRedirectDomain": "vercel.com",
+  "hostRequired": true,
   "records": [
     {
       "type": "TXT",
--- a/vercel.com.website.json
+++ b/vercel.com.website.json
@@ -3,7 +3,7 @@
   "providerName": "Vercel",
   "serviceId": "website",
   "serviceName": "Vercel Websites",
-  "version": 3,
+  "version": 4,
   "description": "Enables a domain to work with Vercel",
   "syncPubKeyDomain": "domainconnect.vercel.com",
   "syncRedirectDomain": "vercel.com",
@@ -34,6 +34,13 @@
       "groupId": "apex",
       "host": "@",
       "pointsTo": "%ip%",
+      "ttl": 600
+    },
+    {
+      "type": "CNAME",
+      "groupId": "apex-cname",
+      "host": "@",
+      "pointsTo": "%apex-cname%",
       "ttl": 600
     }
   ]


vercel.com.website.json
```diff
@@ -3,10 +3,11 @@
   "providerName": "Vercel",
   "serviceId": "website",
   "serviceName": "Vercel Websites",
-  "version": 3,
+  "version": 4,
   "description": "Enables a domain to work with Vercel",
   "syncPubKeyDomain": "domainconnect.vercel.com",
   "syncRedirectDomain": "vercel.com",
+  "hostRequired": true,
```
elliotdauber, line 10: @pawel-kow is there any way to do this without needing to add hostRequired: true?

elliotdauber: i actually am not totally sure this version is correct because we still want to be able to just do the apex verification (_vercel.APEX_DOMAIN) but with hostRequired: true we may need to now specify a host, which would break that. Can you provide some guidance please?

pawel-kow: If it is about RR in apex-cname group, then just put "_vercel" as host, if I get your idea correct.

elliotdauber: For example we have this other record

```
{
      "type": "TXT",
      "groupId": "apex-verification",
      "host": "_vercel",
      "data": "vc-domain-verify=%apex-verification%",
      "ttl": 600
 },
```

If we have `hostRequired: true`, and i want to apply this record for example.com (TXT _vercel.example.com), how would I do that if I have to specify a host? Because it would go to _vercel.myhost.example.com

elliotdauber: But also for this new record I'm adding in this PR, I explicitly want to use a root CNAME (cloudflare supports this and does CNAME flattening)

elliotdauber: (But the [Online Editor](https://domainconnect.paulonet.eu/dc/free/templateedit) failed validation when I didn't have `hostRequired: true`, which is why I added it)

pawel-kow: the problem is that the repository is for all providers, not just cloudflare - this is why templates with CNAME at apex are not accepted (at least not without hostRequired).
What cloudflare does, however, is that they also ignore hostRequired, so that you can define your template with hostRequired=true and use it with cloudflare on apex (without host parameter). Not clean but works.
To test the template in the online editor you may just use APEXCNAME instead of CNAME. It would be nice if cloudflare would support APEXCNAME and translate to CNAME internally, but they don't, so the workaround is needed.

elliotdauber: Ok thanks for clarifying, understood. This current setup will work for us then. Good on my end to merge.

pawel-kow (approved):

PR #957, opened 2026-04-08
bongo.bot.email: add syncBlock: false

danieltwinbolt:
Adds `"syncBlock": false` to the `bongo.bot.email` template.

## Why

Cloudflare's Domain Connect implementation requires `syncBlock` to be explicitly set to `false` (synchronous flow only) for templates to be onboarded to Cloudflare-hosted DNS zones. Without this field, the BongoBot email template cannot be offered through Cloudflare's Domain Connect UI.

Reference: https://developers.cloudflare.com/dns/reference/domain-connect/

## Change

Single-field addition — no record or signing changes:

```diff
 "version": 1,
+"syncBlock": false,
 "syncPubKeyDomain": "bongo.bot",
```

### DIFF
--- a/bongo.bot.email.json
+++ b/bongo.bot.email.json
@@ -4,6 +4,7 @@
 	"serviceId": "email",
 	"serviceName": "Email Sending & Receiving",
 	"version": 1,
+	"syncBlock": false,
 	"syncPubKeyDomain": "bongo.bot",
 	"syncRedirectDomain": "bongo.bot",
 	"description": "Configure DNS records for sending and receiving email through BongoBot.",


pawel-kow:
@danieltwinbolt syncBlock is false by default and afaik CF does not have any such strange requirements to have an explicit value for this. Many templates in this repo do not state it and are used with CF.
CC @kerolasa

kerolasa:
Short answer: yes, Cloudflare requires `syncBlock: false`

Longer answer. My understanding of the syncBlock is that template does not support the synchronous flow. That is contradiction in Cloudflare scope where synchronous flow is the only type flow we support.

Reference: https://developers.cloudflare.com/dns/reference/domain-connect/#1-add-templates-to-the-repository

pawel-kow:
> Short answer: yes, Cloudflare requires `syncBlock: false`
> 
> Longer answer. My understanding of the syncBlock is that template does not support the synchronous flow. That is contradiction in Cloudflare scope where synchronous flow is the only type flow we support.
> 
> Reference: https://developers.cloudflare.com/dns/reference/domain-connect/#1-add-templates-to-the-repository

@kerolasa citing the reference:
> *If present*, you must set the syncBlock field on your template to false. This means the template flow will be synchronous, which is the only option supported by Cloudflare.

This means that syncBlock can be also missing. So this PR is just pointless.

danieltwinbolt:
Thanks @pawel-kow, you're right — re-reading the Cloudflare docs, the exact wording is *"If present, you must set the syncBlock field on your template to false"*, so omitting it entirely is compliant. Closing as unnecessary. Sorry for the noise!

PR #980, opened 2026-04-13
docs: add AGENTS.md file

kerolasa:
The agents file is supposed to be read by various AI tools.  My attempts is to give a hints in it to Service Providers who are using agents to have greater probability of submitting acceptable merge requests.  Maybe this works, and even if not at least I tried.

Reference: https://agents.md/
CC: Pawel Kowalik <pawel.kowalik@denic.de>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,78 @@
+# AI Agent Guidelines for Domain Connect Templates
+
+This document provides guidance for AI tools working with Domain Connect templates in this repository.
+
+## Scope
+
+These rules apply when **adding or modifying templates** in this repository.
+
+## Template Schema
+
+The template file schema is defined in [`template.schema`](template.schema). This schema **MUST be followed** for all template files.
+
+Key schema requirements:
+- Templates must include required fields: `providerId`, `providerName`, `serviceId`, `serviceName`, `records`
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,77 @@
...
+7. **Scope host variables narrowly** - Fix non-variable parts in host fields
+8. **Never use variables in `host` to target subdomains** - Use the standard `host` parameter instead
+9. **Never write `%host%` explicitly** in the `host` attribute - The protocol handles this automatically
+10. **Set `essential`** on records users may need to change independently
+
+For complete details, see the [README.md](README.md).
+
+## References
+
+- [Domain Connect Specification](https://github.com/Domain-Connect/spec/blob/master/Domain%20Connect%20Spec%20Draft.adoc)
```
pawel-kow: This specification is better I think, especially for agentic use:
https://www.ietf.org/archive/id/draft-ietf-dconn-domainconnect-01.html

pawel-kow (dismissed):
Good idea.

AGENTS.md
kerolasa: The bea779b...b47d8cc updated the spec to .html version.

I alsop added another reference to dc-debug-pubkey.  Sure enough the pubkey TXT record checker is not part of Template repo, but I have had so many service providers with broken records that I want to give them all changes to catch issues with their setups.

PR #1029, opened 2026-04-22
update dc-template-lint to use --merge-or-fail option

kerolasa:
This new option is was created to be used as a single definition that will do all the right review things during review.  If the linter exit value is zero (success) merge can proceed automatically.  Possible messages, including info level ones, will cause linter job to fail and merge to be possible only via human review process.

What needs to be done after this change is merged:

To enable auto-merge for this repository, a repository admin needs to:

1. Go to Settings → General → Pull Requests section in the GitHub repository
2. Enable "Allow auto-merge" checkbox
3. Configure branch protection rules for master branch:
   - Require the workflow jobs (lint, validate_json_schema) to pass before merging
   - Enable "Require merge queue" (since the workflows already support merge_group)

Requested-by: Pawel Kowalik \<pawel.kowalik@denic.de>
Signed-off-by: Sami Kerola \<kerolasa@iki.fi>

### DIFF
--- a/.github/workflows/automerge-check.yml
+++ b/.github/workflows/automerge-check.yml
@@ -1,0 +1,88 @@
+name: Automerge Check
+
+on:
+  pull_request_target:
+    branches:
+      - master
+    types: [opened, synchronize, reopened]
+
+jobs:
+  automerge-check:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+
+    steps:
+      - name: Setup refs
+        id: setup-refs
+        run: |
+          echo "base_sha=${{ github.event.pull_request.base.sha }}" >> "$GITHUB_OUTPUT"
+          echo "head_sha=${{ github.event.pull_request.head.sha }}" >> "$GITHUB_OUTPUT"
+
+      - name: Check out code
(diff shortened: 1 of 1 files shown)


pawel-kow (commented):
My intention was a separate GH action, so that the regular check won't just fail.

Intended logic:
- run linter with merge-or-fail on all changed/added files
- if it fails -> end silently
- if it succeeds
  a) for now - just add a label "automerge-possible"
  b) after a test run if we see there are no false-positives (PRs which we would not like auto-commited that linter does not yet detect) - change the action to make real automatic merge (under condition that other needed checks succeed).

kerolasa:
> My intention was a separate GH action, so that the regular check won't just fail.

Right, my bad. I updated the change entirely. Commit message is now:

---

Add separate automerge-check GitHub Action workflow

This workflow runs independently from the regular lint check and:
- Runs linter with --merge-or-fail on all changed/added JSON files
- Ends silently if linter fails (does not block PR)
- Adds "automerge-possible" label if linter succeeds

This is phase (a) of the automerge implementation. After testing confirms
no false-positives, phase (b) can enable actual automatic merging.

PR #1444, opened 2026-07-21
Add jmjbillings.com ar-sending template (Postmark AR email)

JMJBILLINGS:
# Description

New template for JMJ Billings (providerId jmjbillings.com, serviceId ar-sending). Configures Postmark email sending for a customer's dedicated subdomain (standard `host` parameter, e.g. mail.example.com): DKIM TXT, Return-Path CNAME (pm-bounces -> pm.mtasv.net), and inbound MX (inbound.postmarkapp.com, priority 10). Requests are signed; public key published at _dck1.jmjbillings.com.

## Type of change

Please mark options that are relevant.

- [x] New template
- [ ] Bug fix (non-breaking change which fixes an issue in the template)
- [ ] New feature (non-breaking change which adds functionality to the template)
- [ ] Breaking change (fix or feature that would cause existing template behavior to be not backward compatible)

# How Has This Been Tested?

Please mark the following checks done
- [x] Template functionality checked using [Online Editor](https://domainconnect.paulonet.eu/dc/free/templateedit)
- [x] Template file name follows the pattern `<providerId>.<serviceId>.json`
- [x] resource URL provided with `logoUrl` is actually served by a webserver

# Checklist of common problems

- [x] `syncPubKeyDomain` is set
- [x] `warnPhishing` is **not** set alongside `syncPubKeyDomain`
- [x] `syncRedirectDomain` is set (the synchronous flow uses `redirect_uri`)
- [x] no TXT record contains SPF content
- [x] `txtConflictMatchingMode` — n/a: the DKIM TXT host embeds a per-instance Postmark selector (`%dkimSelector%._domainkey`), so no per-label conflict is possible
- [x] no variable is used as a bare full record value — EXCEPTION with justification: `%dkimValue%` is a DKIM public key record (`k=rsa;p=...`) whose full value is prescribed by the DKIM standard and cannot carry a service prefix; the host is constrained to `%dkimSelector%._domainkey`
- [x] no bare variable is used as the full `host` label (`%dkimSelector%._domainkey` — fixed `._domainkey` suffix)
- [x] no variable is used in the `host` field to create a subdomain — the sending subdomain is supplied via the standard `host` apply parameter (`hostRequired: true`)
- [x] `%host%` does not appear explicitly in any `host` attribute
- [x] `essential` — all three records are integral to the service; none are independently user-modifiable records (no DMARC record in this template)

## Online Editor test results

**Editor test link(s):**
[...]

PR #1446, opened 2026-07-22
Allow fork-PR checkout in pull_request_target workflows

### LINKED ISSUE #1445: CI broken for all fork PRs — actions/checkout refuses pull_request_target checkout
All five pull_request_target workflows (automerge-check, JSON Schema Validation, lint, check-description, label-templates) currently fail within seconds for every fork PR, at the actions/checkout step. Run annotation:

> Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. ... To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

This appears to be a GitHub Actions runner-side policy change. Since pull_request_target runs the base repo's workflow definitions, fork PRs cannot fix or work around it — all contributor PRs are blocked from going green (example: #1444, where all five checks die in 3-10s before any validation runs).

The fix appears to be adding `allow-unsafe-pr-checkout: true` to the actions/checkout step in each affected workflow. The workflows only run base-repo tooling (dc-template-linter, check-jsonschema, .github/scripts) against changed root-level JSON files rather than executing contributed code, so the risk profile seems acceptable — maintainers' call.

I've opened a PR with the five one-line changes for consideration.

JMJBILLINGS:
Fixes #1445.

All five pull_request_target workflows currently fail at the actions/checkout step for every fork PR — the Actions runner now refuses fork-PR checkout in pull_request_target workflows unless the checkout step sets `allow-unsafe-pr-checkout: true` (see any recent fork PR's run annotations, e.g. #1444).

This adds that one line to the checkout step in each of the five workflows. Risk note per https://gh.io/securely-using-pull_request_target: these workflows run base-repo tooling (dc-template-linter, check-jsonschema, .github/scripts) against changed root-level JSON files and do not execute contributed code, so the opt-in appears safe here — but please review.


kerolasa:
Updates in this pull result to no change. Is this pull request obsolete?

JMJBILLINGS:
Yes - upstream CI is working again without these changes (checks have been running on fork PRs for a while now), so this is superseded. Closing - thanks for taking a look!

If you have a moment: the template PR this was originally unblocking, #1444, has been all-green for ~3 weeks with the single review question answered (hostRequired justification - the template's MX must never land on a customer apex). Two DNS providers are waiting on that merge to auto-deploy the template, so a review would be much appreciated.

PR #1479, opened 2026-07-29
add otterdeploy, Connect a custom domain

artzkaizen:
# Description

Adds a template for **otterdeploy**, a self-hostable deployment platform. The service connects a user's domain apex to the public IPv4 address of their own otterdeploy server — a single `A` record, `%IP%` supplied by the user's server.

Template file: `otterdeploy.com.connect.json`

## Type of change

- [x] New template

# How Has This Been Tested?

- [ ] Template functionality checked using [Online Editor](https://domainconnect.paulonet.eu/dc/free/templateedit)

Linter passes with exit code 0:

```
$ dc-template-linter -loglevel error -tolerate info -logos otterdeploy.com.connect.json
$ echo $?
0
```

- [ ] no variable is used as a bare full record value (e.g. `@ TXT "%foo%"`) unless necessary — prefer `@ TXT "service-foo=%foo%"`; if bare, justify in the PR description
- [ ] `essential` is set to `OnApply` on records the end user may need to modify or remove without breaking the template (e.g. DMARC)

### Comments on unticked items

**Bare variable as full record value.** The template's single record is `A @ → %IP%`. An `A` record's value *is* an address, so there is no fixed prefix to scope it with — the guideline's `service-foo=%foo%` pattern has no analogue for address records. The variable is documented in `variableDescription` ("IP is the public IPv4 address of the user's otterdeploy server") and is constrained by the record type. Happy to change this if maintainers prefer a different shape.

**`essential`.** Not set. This template installs exactly one record, which *is* the service — if it is removed the domain no longer resolves to the user's server, so there is no "user may modify this without breaking the template" case that `essential: OnApply` is meant to cover. Please tell me if you would rather see it set explicitly.

**Note on `syncRedirectDomain` for Cloudflare.** Running the linter with `-cloudflare` reports an informational `DCTL5003: syncRedirectDomain is not supported`. Retained because the checklist above requires it for the synchronous flow; noting it here for transparency since Cloudflare is an intended DNS provider for this template.

## Online Editor test results

**Editor test link(s):** 

> ⚠️ **Not yet supplied — this PR is a draft for exactly this reason.**
>
> Per `AGENTS.md`, these links must be produced by a human actually using the Online Editor and must never be fabricated. Two runs are still required before this PR is marked ready for review:
>
> - [ ] one test against a domain apex
> - [ ] one test against a nested subdomain using the `host` parameter
>
> The `Copy Markdown` output from each will be pasted here, and the "Template functionality checked using Online Editor" box above ticked, before this leaves draft.

### DIFF
--- a/otterdeploy.com.connect.json
+++ b/otterdeploy.com.connect.json
@@ -1,0 +1,21 @@
+{
+  "providerId": "otterdeploy.com",
+  "providerName": "otterdeploy",
+  "serviceId": "connect",
+  "serviceName": "Connect a custom domain",
+  "version": 1,
+  "syncBlock": false,
+  "syncPubKeyDomain": "domainconnect.otterdeploy.com",
+  "syncRedirectDomain": "get.otterdeploy.com",
+  "logoUrl": "https://otterdeploy.com/favicon.svg",
+  "description": "Connect a custom domain to an otterdeploy server.",
+  "variableDescription": "IP is the public IPv4 address of the user's otterdeploy server.",
+  "records": [
+    {
+      "type": "A",
+      "host": "@",
+      "pointsTo": "%IP%",
+      "ttl": 300
+    }
+  ]
+}


artzkaizen:
Closing in favour of #1480, which is the identical commit (`b4d43c5`) opened from the `otterdeploy` organisation fork rather than a personal one — the org is the actual owner of the `otterdeploy.com` provider ID this template claims. Apologies for the noise.

PR #1808, opened 2026-09-10
Add FadeHost templates: minecraft-server (SRV) and community-website (CNAME)

bernisnukic:
# Description

Two new templates for FadeHost (fadehost.com), a Minecraft server host:

- `fadehost.com.minecraft-server.json`: one SRV record so players can join a Minecraft Java Edition server through a subdomain of the customer's own domain (`_minecraft._tcp.<host>.<domain>` → `%node%.fadehost.net.` on `%port%`). The target variable can only fill the node label; the `.fadehost.net.` suffix is fixed in the template.
- `fadehost.com.community-website.json`: one CNAME so a FadeHost community website (join page, live map, leaderboard) is served on a subdomain of the customer's domain (`<host>.<domain>` CNAME `join.fadehost.net`). No variables.

Both templates are `hostRequired` (subdomain only), signed (`syncPubKeyDomain`: fadehost.com, key published at `_dcpubkeyv1.fadehost.com`) and return the browser to `laplace.fadehost.com` (`syncRedirectDomain`). The logo at `https://fadehost.com/logo.svg` is live.

## Type of change

Please mark options that are relevant.

- [x] New template
- [ ] Bug fix (non-breaking change which fixes an issue in the template)
- [ ] New feature (non-breaking change which adds functionality to the template)
- [ ] Breaking change (fix or feature that would cause existing template behavior to be not backward compatible)

# How Has This Been Tested?

Please mark the following checks done
- [x] Template functionality checked using [Online Editor](https://domainconnect.paulonet.eu/dc/free/templateedit)
- [x] Template file name follows the pattern `<providerId>.<serviceId>.json`
- [x] resource URL provided with `logoUrl` is actually served by a webserver

# Checklist of common problems

Mark all the checkboxes after conducting the check. Comment on any point which is not fulfilled.
See [Template Quality Guidelines](../README.md#template-quality-guidelines) for details and rationale on each rule.

- [x] `syncPubKeyDomain` is set — **this is mandatory**; omitting it requires explicit justification in the PR description or the PR will be rejected
- [x] `warnPhishing` is **not** set alongside `syncPubKeyDomain` — the two must not appear together
- [x] `syncRedirectDomain` is set whenever the template uses `redirect_uri` in the synchronous flow
- [x] no TXT record contains SPF content (`"v=spf1 ..."`) — use the `SPFM` record type instead
- [x] `txtConflictMatchingMode` is set on every TXT record that must be unique per label or content prefix (e.g. DMARC)
[...]

PR #1901, opened 2026-09-17
Add datascale.kr.gameshield template

sirin-interact:
# Description

New template for datascale GameShield, a Minecraft DDoS protection proxy. A customer points a domain they own at their proxy: an A record for the address players type, and an SRV record so they do not have to type the port. Our control plane hands every proxy an arbitrary port on a shared IP, so without the SRV record players would have to type `example.com:51473`.

Both records are needed together, and the template is meant to work on the apex as well as on a subdomain — most of our customers put their server on the domain itself.

## Type of change

- [x] New template

# How Has This Been Tested?

- [x] no variable is used as a bare full record value (e.g. `@ TXT "%foo%"`) unless necessary — prefer `@ TXT "service-foo=%foo%"`; if bare, justify in the PR description
- [x] `essential` is set to `OnApply` on records the end user may need to modify or remove without breaking the template (e.g. DMARC)

Notes on the checklist:

- **`redirect_uri`, TXT records:** the template uses neither, so `syncRedirectDomain` and `txtConflictMatchingMode` do not apply.
- **`%ip%` is a bare record value.** It is the `pointsTo` of an A record, which can only ever be an address — there is no fixed prefix to wrap it in.
- **`%target%` is not bare.** The SRV target is `%target%.datascale.kr`, so the variable can only ever resolve inside our own zone. We pass the label without the suffix.
- **`%port%` is the whole `port` value,** which the specification requires for numeric fields.
- **`essential` is left at its default (`Always`).** Both records *are* the service — if the customer removes either one, players cannot reach their server, so there is no record here that should survive being changed independently.
- **No `hostRequired`.** The apex is the main case for us: customers type their own domain into our console, and most of them use the domain itself rather than a subdomain. The A record (not CNAME) is chosen precisely so the apex works.

## Online Editor test results

**Editor test link(s):** 
[Test datascale.kr/gameshield example.com/play](https://domainconnect.paulonet.eu/dc/free/templateedit?token=H4sIACL4q2oC%2F%2B1VbW%2FaMBD%2BK5alfmoISUiBZppUtE5bO6lD68taIRQZ24C1EKe2E5oh%2FvvOgQBh7T5P2r7Fd8%2F57p57fFlhwxdZQgzH0QpnShaCcXXFcIQZMURTknD3h8LOzndDFvzQCy7NVSEor6Jm4NZzwRO2dxyHoE9guK1BBVdayBRHvoMTOZP3KgHw3JhMR%2B32YRVtQWXq6mIGUYxrqkRmqkg8lCI1iCAmF0SkiFTfr6RD0MVLGSGSogFSnErF0FQqZOYcEcYU1xoBGyWUhEyZcQeQzKJvvz3UeC0tvIRcKJUGzUnBkZEVvLonk8q4ti%2BiBJkk%2FLJR6YnITiJ0NSzCXUI5reKOq3yHTuxVgL77MKxuRURrMUs5q%2FLZVFucIWrGLXJnRHOpDcpTGBg6pNBBS2HmMjcV1D10IZ1Pp%2BLFjq1M6TCffOHlZcXn72LYUKFxNFph2zggBmC2SeHzwsrFTkTfyW3LYDHGzrXjeXjt7KKAV3ClG4FcNFBWckIqYUow2OOSi9ncbA%2BWD3t3RdFGnkZSaYNjQ7O9%2BKxhIVJOFZkam6DiyoZuWWuQgNfjtYN%2FypTH%2Bx7HILeaCP5C4L1wl8rFvmErGTjNlMyzWNQxGVFkoe27qqNHjfBxHT%2FaXABnkdmT3z13%2FaDvBkHX9T1rr4veIF3iTwLaYaF1bYgY4eDsrHuGbfUgEal4bKVCTK6AAaNy7uBFnhgRkyXZmxiNSZYlJTSrwQv3NAe6Hcy2P0sUnI7K284MRubgmFHbr7CboO%2F1Ot3Am7YY56wVnk9pq09Y2Jr4hJz1Asppv3OwV17bOW8ulibv8Ih4agSxwx8kS1JqvH5LY81WGmy6R9n3bR2I64%2Fa2gvW28vVq2dUTeiv4mjsgE7%2FT%2FxfmjhsDA1%2FLBYTCw28oNvyzlt%2B784PoyCMOqEbhJ3AD089L6qWsOHabLhbwW453Cr4SXwlT9%2BFHvYGj91u8cg%2B3z%2F3F6dLGraf027b%2B3Lz0LvOi%2Fbp9cf3eP0LPm8WYGsIAAA%3D)

**The apex test could not be produced — the editor rejects any apex SRV**, with `Invalid data for SRV host: __tcp.@`. I tried all three combinations: empty host with `"name": "@"`, empty host with `"name": ""`, and host `@` with `"name": "@"`. All three fail with the identical message, and the `service` value never appears in it — with a host set, the same template composes `_minecraft._tcp.play` correctly, so the field is only dropped on the apex path. The same limitation was reported on #1808 ("the apex variant … could not be applied in the online editor"), where it was resolved by scoping that template to subdomains.

I did not set `hostRequired` here, because the apex is exactly the case this template exists for, and the records themselves are valid at the apex (`_minecraft._tcp.example.com` SRV plus an apex A record). If you would rather have the template scoped to subdomains until the editor handles apex SRV, say the word and I will add `hostRequired: true`.

Other checks run locally:

```
dc-template-linter -loglevel info -logos -merge-or-fail datascale.kr.gameshield.json   # clean, exit 0
dc-template-linter -cloudflare datascale.kr.gameshield.json                            # clean, exit 0
dc-debug-pubkey -dns _dcpubkeyv1.datascale.kr                                          # "record is ok"
dc-debug-pubkey -key _dcpubkeyv1 datascale.kr.gameshield.json <private key> '<query>'  # signs and verifies
```

The file is formatted with the linter's `-pretty -indent 4` output.


sirin-interact:
The description checker is asking for the apex test, so here is what I found when I tried to produce it — I think the editor has a bug on the apex SRV path.

I probed it with placeholder values (`"service": "_svc"`, `"protocol": "_prot"`, `"name": "@"`, no host):

```
Invalid data for SRV host: __prot.@
```

The composed host is `"_" + protocol + "." + name`: `service` is dropped entirely and a literal `_` is put in its place. That is why every apex attempt reports `__tcp.@` regardless of what `service` contains — `_minecraft` and `minecraft` both give the identical message. With a host set, the same template composes `_minecraft._tcp.play` correctly, so only the apex path is affected.

Tried and failing identically: empty host with `"name": "@"`, empty host with `"name": ""`, and host `@` with `"name": "@"`. The same symptom was reported on #1808.

So I cannot attach an apex link for this template as submitted. The records themselves are valid at the apex (`_minecraft._tcp.example.com` SRV alongside an apex A record), and the apex is the case this template exists for — our customers type their own domain into our console and most use the domain itself, which is also why the first record is an A and not a CNAME.

Happy to go either way: leave it as is if you are comfortable reviewing the apex without an editor link, or I will add `hostRequired: true` and scope it to subdomains like #1808 did. Just tell me which you prefer.

